PR #36, opened 2021-06-24
unify build process for different workflows

wkozyra95:
# Why

We want to remove the need for workflows.

# How

- merge job definitions
- remove ManagedBuildContext
- merge builder functions
- simplify eject provider
  - do not remove android/ios dirs
  - do not create git repo if it does not exists

# Test Plan

run android builds
TODO ios


dsokal (approved):
👍

PR #45, opened 2021-09-07
update expo-cli and config-plugins

wkozyra95:
# Why

latest cli properly handles case where name consists only of unicode character

# How

# Test Plan

run android build


dsokal (approved):
🚀

PR #62, opened 2021-12-09
support partial builds

wkozyra95:
# Why

Support partial builds that can be used to implement inspect functionality in eas-cli

# How

# Test Plan

run build and build:inspect with various options

### DIFF
--- a/packages/local-build-plugin/src/exit.ts
+++ b/packages/local-build-plugin/src/exit.ts
@@ -4,14 +4,25 @@ const handlers: (() => void | Promise<void>)[] = [];
 let shouldExitStatus = false;
 
 export function listenForInterrupts(): void {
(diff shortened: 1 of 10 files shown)


packages/local-build-plugin/src/exit.ts
```diff
@@ -4,7 +4,12 @@ const handlers: (() => void | Promise<void>)[] = [];
 let shouldExitStatus = false;
 
 export function listenForInterrupts(): void {
+  let handlerCalled = false;
   const handleExit = async (): Promise<void> => {
+    if (handlerCalled) {
```
wkozyra95, line 9: not sure why, but if I call `childProcess.kill()` local build receives to signals

dsokal (approved):

PR #75, opened 2022-02-10
[build-tools] store release channel name as a string resource

jkhales:
# Why

Android makes a best guess cast when interpreting metadata values. This can lead to discordance between the releaseChannel stored by the build and the releaseChannel of a published update. 

The motivating example/bug is when a user had `1.0` as a releaseChannel.

# How

Store the releaseChannel as a string resource to ensure Android interprets it as a string.

# Test Plan

1. set releaseChannel to `1.0` in eas.json
2. build locally, install into an emulator.
3. publish to `1.0`
4. confirm app downloads and loads update.

### DIFF
--- a/packages/build-tools/src/android/__tests__/expoUpdates.test.ts
+++ b/packages/build-tools/src/android/__tests__/expoUpdates.test.ts
@@ -47,7 +47,7 @@ describe(androidSetClassicReleaseChannelNativelyAsync, () => {
   test('sets the release channel', async () => {
     const reactNativeProjectDirectory = fs.mkdtempSync('/expo-project-');
     fs.ensureDirSync(reactNativeProjectDirectory);
-    const releaseChannel = 'default';
+    const releaseChannel = 'blah';
     const ctx = {
       reactNativeProjectDirectory,
       job: { releaseChannel },
@@ -75,7 +75,15 @@ describe(androidSetClassicReleaseChannelNativelyAsync, () => {
         newAndroidManifest,
         AndroidMetadataName.RELEASE_CHANNEL
       )
-    ).toBe(releaseChannel);
+    ).toBe('@string/release_channel');
+
+    const stringResourcePath = await AndroidConfig.Strings.getProjectStringsXMLPathAsync(
+      reactNativeProjectDirectory
+    );
+    const stringResourceObject = await AndroidConfig.Resources.readResourcesXMLAsync({
+      path: stringResourcePath,
+    });
+    expect((stringResourceObject.resources.string as any)[0]['_']).toBe(releaseChannel);
   });
 });
 describe(androidSetChannelNativelyAsync, () => {
--- a/packages/build-tools/src/android/expoUpdates.ts
+++ b/packages/build-tools/src/android/expoUpdates.ts
@@ -45,9 +45,7 @@ export async function androidSetClassicReleaseChannelNativelyAsync(
   ctx: BuildContext<Job>
 ): Promise<void> {
   const { releaseChannel } = ctx.job;
-  if (!releaseChannel) {
-    throw Error('releaseChannel must be defined');
-  }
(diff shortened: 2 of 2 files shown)


packages/build-tools/src/android/__tests__/expoUpdates.test.ts
```diff
@@ -47,7 +47,7 @@ describe(androidSetClassicReleaseChannelNativelyAsync, () => {
   test('sets the release channel', async () => {
     const reactNativeProjectDirectory = fs.mkdtempSync('/expo-project-');
     fs.ensureDirSync(reactNativeProjectDirectory);
-    const releaseChannel = 'default';
+    const releaseChannel = 'blah';
```
kbrandwijk, line 50: Can we also include a test for this specific case? So with a releaseChannel that would previously be incorrectly 'cast'?

jkhales: This test checks that the meta data is stored as a "string resource".

A string resource is guaranteed to be interpreted as a string.

wkozyra95: we can't really test it because the resulting interpreted value is available in android runtime

packages/build-tools/src/android/expoUpdates.ts
```diff
@@ -44,7 +44,11 @@ export async function androidSetChannelNativelyAsync(ctx: BuildContext<Job>): Pr
 export async function androidSetClassicReleaseChannelNativelyAsync(
   ctx: BuildContext<Job>
 ): Promise<void> {
-  assert(ctx.job.releaseChannel, 'releaseChannel must be defined');
+  const { releaseChannel } = ctx.job;
+  if (!releaseChannel) {
+    throw Error('releaseChannel must be defined');
+  }
```
wkozyra95, line 50:
```suggestion
  const { releaseChannel } = ctx.job
  assert(releaseChannel, 'releaseChannel must be defined');
```

It's just a semantic difference, but assert should've been used to guard against programmer error and exceptions from user errors/external issues. Here if releaseChannel is null then it's fault of the code that is calling this function

packages/build-tools/src/android/expoUpdates.ts
```diff
@@ -53,12 +57,30 @@ export async function androidSetClassicReleaseChannelNativelyAsync(
...
+  );
+  const stringResourceObject = await AndroidConfig.Resources.readResourcesXMLAsync({
+    path: stringResourcePath,
+  });
+  const resourceName = 'release_channel';
+  const releaseChannelResourceItem = AndroidConfig.Resources.buildResourceItem({
+    name: resourceName,
+    value: escapedReleaseChannel,
+  });
+  const newStringResourceObject = AndroidConfig.Strings.setStringItem(
```
wkozyra95, line 72: will this work for multi-flavor bare projects that do not have `app/src/main/res` directory? It's a very niche use-case so I'm not saying to support it, but we should at least throw some meaningful error message.

Does it work if release_channel is already specified?

wkozyra95 (approved):

packages/build-tools/src/android/expoUpdates.ts
wkozyra95: verified on my own
- it works even if `res` directory does not exists
- if the release channel is already present it's overridden

jkhales: Good questions and 🙏  thank you for investigating!

This seems like desirable behavior to me.

packages/build-tools/src/android/__tests__/expoUpdates.test.ts
kbrandwijk: Ah, I misread the way the result is validated. Thanks for the explanation, that makes perfect sense then.

PR #96, opened 2022-05-05
support pnpm and fix npm workspaces

### LINKED ISSUE #90: npm v7+ monorepos are broken in EAS
I know we don't "officially support" npm workspaces, but would be great if it worked. One of the big issues with npm v7+ for React Native is the peer dependencies. We often encounter these peer deps issues in React Native, but we have no way of turning this off in EAS build.

## Issue

What's going on is that monorepos with npm v7 needs to be installed (`npm install`) from the root of the monorepo. When running `npm install` inside the **apps/my-app**, it fails because it can't find the right monorepo packages.

A weird behavior I found is, you can actually turn this off [with the .npmrc file](https://docs.npmjs.com/cli/v8/using-npm/config#workspaces). But adding this to the root of the monorepo, or globally in the worker, still doesnt get picked up when running `npm install` inside the **apps/my-app** folder.

## Proposed fix

Instead of calling `<packageManager> install` on [this line](https://github.com/expo/eas-build/blob/3383c40fad8454a4ab4017f8c6997bf3e9d99432/packages/build-tools/src/utils/prebuild.ts#L24), we might want to rely on package manager to resolve to the right monorepo root instead. That, combined with the "find workspace root" method [I RFC'd in this PR](https://github.com/expo/expo-cli/pull/4311) should get us support for all package managers in `@expo/package-manager`.

One thing I don't necessarily like is that we force people to use one of the package manager supported managers. If we have a `installCommand` in EAS json, we could let the user decide however they want to install the node modules inside the project. But that's just a pet peeve 🤷 

## Example of failures

- npm v7 - https://expo.dev/accounts/bycedric/projects/eas-benchmark-npm-v7
- npm v8 - https://expo.dev/accounts/bycedric/projects/eas-benchmark-npm-v8

wkozyra95:
# Why

Closes #90 

Adds support for pnpm
Fixes support for npm workspaces

# How

- run install from the root of workspace
- do not pass CI and EAS_BUILD envs to install scripts to make sure it does not trigger immutable installs
- unify installing dependencies before and after prebuild

## Breaking changes

- when there are both yarn and npm lockfile the npm takes precedence (previously it was yarn)
- if there is no lockfile use yarn. Currently, we default to npm if there is no lockfile but to yarn if there is an error detecting package manager, so unifying would simplify the situation and yarn is probably the least problematic (e.g. issues with peer dependencies)

# Test Plan

run builds in project without workspaces on all 3 package managers, pnpm build failed but it looks like everything was called correctly

cc @byCedric do you have any pnpm or npm@7 project with workspaces that I could test it on?


byCedric:
@wkozyra95 Yeah I have!
- https://github.com/byCedric/eas-monorepo-example/tree/test-pnpm-sdk44 → if you want to test a monorepo
- https://github.com/byCedric/expo-monorepo-benchmark/tree/main/pnpm-v7 → same thing, but with versions v6 and v7

Creating a single app (non-monorepo) shouldn't be hard either, [see these steps](https://github.com/expo/expo-cli/pull/4311#issue-1206113638) under "Want to try it out?".

wkozyra95:
@brentvatne do you have any opinion about those breaking changes?

dsokal (approved):

byCedric:
As discussed, let's keep the same behavior in EAS (avoiding the breaking change), and make sure the CLI/package manager does the same thing.

brentvatne:
> @brentvatne do you have any opinion about those breaking changes?

it sounds like only one of the two breaking changes remains ("when there are both yarn and npm lockfile the npm takes precedence (previously it was yarn)" being resolved in https://github.com/expo/expo-cli/pull/4369)

> if there is no lockfile use yarn. Currently, we default to npm if there is no lockfile but to yarn if there is an error detecting package manager, so unifying would simplify the situation and yarn is probably the least problematic (e.g. issues with peer dependencies)

this sounds good to me. if you have no lockfile you're asking for trouble and yarn does indeed minimize the risk.

brentvatne (approved):

PR #98, opened 2022-05-13
[license] Automatically update the BSL license's change date

ide:
Why
---
We want to keep the BSL license's change date to be three years after each code commit. It is nicer to have this automated.

How
---
Add a GitHub Action workflow that is set to run daily. It compares the commit time (not author time) of the last commit that modified a file other than LICENSE with the current change date, taking the three year period into account.

See https://github.com/actions/checkout#push-a-commit-using-the-built-in-token for docs on how to push using the built-in GitHub token.

We could also move the script into a .sh file, which is easier to work on. I don't think it matters too much right now, though.

Test Plan
---
Add a new workflow with:
```
on:
  pull_request
```
so that we can test it instead of relying on the daily cron job. After committing this, we will also be able to verify that it runs on the cron schedule.

The expected output in the GHA logs is for `git show` to show a commit that updates the license's date. https://github.com/expo/eas-build/runs/6416935723?check_suite_focus=true


.github/workflows/license.yml
```diff
@@ -0,0 +1,38 @@
+name: License
+on:
+  pull_request:
+    branches: [main]
```
ide, line 4: We don't need to, but we should remove this before merging

dsokal (approved):

.github/workflows/license.yml
```diff
@@ -0,0 +1,38 @@
...
+          commit_date=$(git log -1 --format=%cd --date=short -- . ':!/LICENSE')
+          [ -z "$commit_date" ] && exit 0
+
+          # The date in the license
+          change_date=$(sed --quiet --regexp-extended 's/^Change Date:\s+([0-9]{4}-[0-9]{2}-[0-9]{2})/\1/p' < LICENSE)
+          [ -z "$change_date" ] && exit 0
+          
+          new_change_date=$(date --date "$commit_date +3 years" --iso-8601)
+
+          if [ "$(date --date "$new_change_date" +%s)" -gt "$(date --date "$change_date" +%s)" ]; then
```
wkozyra95, line 29: wouldn't this spam date update after every change?

wkozyra95 (commented):
btw we already have that automated 
- https://github.com/expo/eas-build/blob/main/scripts/bumpLicenseChangeDate.sh
- https://github.com/expo/eas-build/blob/main/scripts/publish.js#L86
If you want to do that on CI remove that first otherwise both solution will interfere with each other

Our current script is running on every publish and if the date is too soon we are updating. We are using different times for comparison and different for update, so the date is updated every 3 months

.github/workflows/license.yml
ide: In my proposal, the cron job would run once per day. The job will update the LICENSE only if there were commits earlier that day. If you feel this is too noisy, we can change the cron schedule to be only once per week or once per month.

ide:
I didn't realize there was already a script running when we publish packages.

What would you and @dsokal prefer? The publish script or a CI job?

wkozyra95:
> I didn't realize there was already a script running when we publish packages.
> 
> What would you and @dsokal prefer? The publish script or a CI job?

CI approach is definitely better, we did not do that initially, so we don't need to deal with committing back from ci job. I would prefer if that job either run our script or the current implementation was modified so we don't commit more often than e.g. every 3 months.

ide:
OK. I deleted the existing script and schedule the CI workflow to run every three months. The cron schedule is current set to 1 day just so we can test it end-to-end. Once we confirm it is working as expected, I will change the cron schedule.

wkozyra95 (approved):

wkozyra95:
@ide It looks like this job is bumping date even if the last change was also a date update

ide:
Yes, I removed that check because I expect we would have at least one non-LICENSE change every three months. We can add it back it if needed but I thought it was better not to have extra unnecessary code.

PR #102, opened 2022-05-14
[local-build-plugin] Update Expo CLI

brentvatne:
# Why

To align with https://github.com/expo/turtle-v2/pull/957


wkozyra95:
already updated in PT that add pnpm support

PR #108, opened 2022-05-26
detect and report corupted packages

wkozyra95:
# Why

Occasionally I have seen reports where the error suggests that there is a problem with corrupted packages in npm caches. One recent issue I was able to reproduce, in that case file returned by verdaccio was correct, but the checksum reported by verdacio API was not.

I'm adding this monitoring to help with monitoring the issues, and as way to provide quick workaround for users that hit those issues 

# How

- detect error based on logs and throw custom error
- add onError callback to runBuildPhase
- report erro to sentry

# Test Plan

will test on turtle


packages/eas-build-job/src/errors.ts
```diff
@@ -124,8 +125,8 @@ export class IncompatiblePodsManagedWorkflowError extends UserError {
 You are seeing this error because either:
   - Versions in the Podfile.lock cached by EAS do not match required values for some of the libraries, it can be triggered when upgrading Expo SDK or any other library with native code. To fix that ${
     usingDefaultCacheConfig
-      ? 'add the "cache.key" field (it can be set to any value) in eas.json to invalidate the cache.'
-      : 'update value of the "cache.key" field in eas.json to invalidate the cache.'
+      ? 'add the "cache.key" field (it can be set to any value) in the build profile in eas.json to invalidate the cache.'
```
wkozyra95, line 128: unrelated, but i saw recently someone that tried add those keys in root of eas.json

dsokal (approved):

PR #142, opened 2022-09-09
Improve error detection and classification

wkozyra95:
# Why

Improve error detection

# How

- Keep 2 lists of error handlers, one is matching errors that we want to show some mitigation steps to users and the second that just assigns error code that we can see in sentry.
- I did not change anything in definitions in packages/build-tools/src/buildErrors/userErrorHandlers.ts , just merged existing error descriptions with regexps that match them in one list.
- Remove onError callback on build phase.

# Test Plan

Add some tests


packages/build-tools/src/buildErrors/buildErrorHandlers.ts
```diff
@@ -0,0 +1,100 @@
+import { BuildPhase, Platform } from '@expo/eas-build-job';
+import escapeRegExp from 'lodash/escapeRegExp';
+
```
wkozyra95, line 3: For now, it's mostly cache errors and a few I saw a lot in sentry. When this will go live I will have a better way to find new unknown errors, and make a second pass at adding unknown errors.

packages/build-tools/src/buildErrors/buildErrorHandlers.ts
```diff
@@ -0,0 +1,100 @@
...
+    // Adding spec repo `24-repository-cocoapods-proxy` with CDN `http://10.254.24.7:8081/repository/cocoapods-proxy/`
+    // [!] No podspec exists at path `/Users/expo/.cocoapods/repos/24-repository-cocoapods-proxy/Specs/1/9/2/libwebp/1.2.0/libwebp.podspec.json`.
+    regexp: /Adding spec repo .* with CDN .*\n\s*\[!\] No podspec exists at path `(.*)`/,
+    // Some pods are hosted on git registries that are not supported e.g. chromium.googlesource.com
+    createError: (match: RegExpMatchArray) =>
+      new TrackedBuildError(
+        'COCOAPODS_CACHE_INCOMPATIBLE_REPO_ERROR',
+        `cocoapods: missing podspec ${match[1]}`
+      ),
+  },
```
szdziedzic, line 65: Can we split this error into two separate (for generic and managed workflow) as you did for `EAS_BUILD_INCOMPATIBLE_PODS` error in `userErrorHandlers`?

wkozyra95: EAS_BUILD_INCOMPATIBLE_PODS needed a different message, this is not the case here. I would prefer to add new field to the sentry error instead

dsokal (approved):

PR #143, opened 2022-09-15
[build-tools] use CocoaPods plugin to handle setting proxy as a source

szdziedzic:
# Why

Our Cocoapods Cache instance can't handle fetching some pods, because of where they are stored.
https://issues.sonatype.org/browse/NEXUS-28026.

This plugin dynamically changes the behavior of CocoaPods. It uses official CDN to download some unsupported by our cache instance pods, even if the cache instance is set to be the source. Also, it dynamically sets the source to be our cache URL if the environment variable is present.

# How

Create a plugin that overrides the behavior of the CDNSource class of CocoaPods when fetching the podspecs and automatically detects if it should set our cache URL as a source depending on the env variable.

Remove the code that modifies `Podfile` depending if the proxy env variable is present.

# Test Plan

Test on staging


wkozyra95 (approved):

dsokal (approved):

packages/build-tools/src/ios/pod.ts
```diff
@@ -9,32 +8,16 @@ import { BuildContext } from '../context';
...
-
   await spawn('pod', ['install'], {
     cwd: iosDir,
     logger: ctx.logger,
-    env: { ...ctx.env, LANG: 'en_US.UTF-8' },
+    env: {
+      ...ctx.env,
+      LANG: 'en_US.UTF-8',
+      ...(ctx.env.EAS_BUILD_COCOAPODS_CACHE_URL
+        ? { COCCOAPODS_CACHE_URL: ctx.env.EAS_BUILD_COCOAPODS_CACHE_URL }
```
wkozyra95, line 18:
```suggestion
        ? { COCOAPODS_CACHE_URL: ctx.env.EAS_BUILD_COCOAPODS_CACHE_URL }
```

wkozyra95 (approved):

szdziedzic:
@wkozyra95 I tested it locally and it seems to work fine :)

randomhajile (approved):

packages/cocoapods-cache-plugin/cocoapods-cache-plugin.gemspec
```diff
@@ -0,0 +1,23 @@
...
+lib = File.expand_path('../lib', __FILE__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require 'cocoapods-cache-plugin/gem_version.rb'
+
+Gem::Specification.new do |spec|
+  spec.name          = 'cocoapods-cache-plugin'
+  spec.version       = CocoapodsCachePlugin::VERSION
+  spec.authors       = ['Expo']
+  spec.email         = ['support@expo.dev']
+  spec.description   = "CocoaPods plugin, which handles fetching the pods which are not supported by Nexus3 cocoapods-proxy repository."
```
wkozyra95, line 11:
```suggestion
  spec.description   = "CocoaPods Nexus plugin overrides the official CDN with the address of the proxy instance. Pods not supported by Nexus will be fetched from the upstream repository."
```

packages/cocoapods-cache-plugin/lib/cocoapods_plugin.rb
```diff
@@ -0,0 +1,79 @@
...
+module Pod
+  class Installer
+    class Analyzer
+      @@_was_using_cocoapods_cache_printed = false
+
+      alias_method :orig_sources, :sources
+
+      # add our own source to the list of sources
+      def sources
+        if COCOAPODS_CACHE_URL
```
wkozyra95, line 18:
```suggestion
        if NEXUS_INSTANCE_URL
```

packages/cocoapods-cache-plugin/lib/cocoapods_plugin.rb
```diff
@@ -0,0 +1,79 @@
...
+          if not @@_was_using_cocoapods_cache_printed
+            puts "Using CocoaPods cache: #{COCOAPODS_CACHE_URL}"
+            @@_was_using_cocoapods_cache_printed = true
+          end
```
wkozyra95, line 22:
```suggestion
          puts "Using CocoaPods cache: #{COCOAPODS_CACHE_URL}" unless @@_was_using_cocoapods_cache_printed
          @@_was_using_cocoapods_cache_printed = true
```

packages/cocoapods-cache-plugin/lib/cocoapods_plugin.rb
szdziedzic: Maybe 
```suggestion
        if NEXUS_COCOAPODS_REPO_URL
```
?
In my opinion, `NEXUS_INSTANCE_URL` name may encourage people to use `http://10.254.24.6` URL vs `http://10.254.24.6/repositories/cocoapods-repo`. What do you think?

wkozyra95 (approved):

PR #145, opened 2022-09-15
[eas-build-job] add new iOS builder image

szdziedzic:
# Why

Add new iOS builder image for XCode 14

# How

Add new iOS builder image for XCode 14

# Test Plan

Tests

### DIFF
--- a/packages/eas-build-job/src/ios.ts
+++ b/packages/eas-build-job/src/ios.ts
@@ -50,6 +50,7 @@ export const builderBaseImages = [
   'macos-monterey-12.1-xcode-13.2',
   'macos-monterey-12.3-xcode-13.3',
   'macos-monterey-12.4-xcode-13.4',
+  'macos-monterey-12.6-xcode-14.0',
 ] as const;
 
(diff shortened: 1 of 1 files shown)


packages/build-tools/src/ios/pod.ts
```diff
@@ -9,28 +8,6 @@ import { BuildContext } from '../context';
 export async function installPods<TJob extends Ios.Job>(ctx: BuildContext<TJob>): Promise<void> {
   const iosDir = path.join(ctx.reactNativeProjectDirectory, 'ios');
 
-  if (ctx.env.EAS_BUILD_COCOAPODS_CACHE_URL) {
```
wkozyra95, line 12: Do not include commits from other reviews, like that. If you want to have stacked PR then this PR should point to the branch where base commit is from, but in this case there is no need for that because changes are not related to each other

wkozyra95 (approved):

packages/build-tools/src/ios/pod.ts
szdziedzic: Sorry, I accidentally did it, it wasn't intentional action

dsokal (approved):

PR #150, opened 2022-09-22
[build-tools] Update error message for pods incompatibility

wkozyra95:
# Why

https://linear.app/expo/issue/ENG-6412/clarify-error-message-for-mismatched-pod-locks

# How

- We already have a workflow tag in sentry, so there is no point in separating errors for managed/bare.
- Suggest using website or --clear-cache options instead of cache key.

# Test Plan

### DIFF
--- a/packages/build-tools/src/buildErrors/userErrorHandlers.ts
+++ b/packages/build-tools/src/buildErrors/userErrorHandlers.ts
@@ -112,33 +112,15 @@ export const userErrorHandlers: ErrorHandler<UserFacingError>[] = [
     //  * changed the constraints of dependency `Firebase/Core` inside your development pod `EXFirebaseCore`.
     //    You should run `pod update Firebase/Core` to apply changes you've made.
     regexp: /CocoaPods could not find compatible versions for pod /,
-    createError: (_, { job }) => {
-      const usingDefaultCacheConfig = job.cache.key === '' || !job.cache.key;
-      return job.type === Workflow.MANAGED
-        ? new UserFacingError(
-            'EAS_BUILD_INCOMPATIBLE_PODS_MANAGED_WORKFLOW_ERROR',
-            `Compatible versions of some pods could not be resolved.
+    createError: () => {
+      return new UserFacingError(
+        'EAS_BUILD_INCOMPATIBLE_PODS_ERROR',
+        `Compatible versions of some pods could not be resolved.
 You are seeing this error because either:
-  - Versions in the Podfile.lock cached by EAS do not match required values in Podspecs of some of the libraries. To fix that ${
-    usingDefaultCacheConfig
-      ? 'add the "cache.key" field (it can be set to any value) in the build profile in eas.json to invalidate the cache.'
-      : 'update value of the "cache.key" field in the build profile in eas.json to invalidate the cache.'
-  }
+  - Versions in the Podfile.lock cached by EAS do not match required values in Podspecs of some of the libraries. To fix that you can re-run build command with "--clear-cache" option, or select "Clear cache and retry build" on the build page.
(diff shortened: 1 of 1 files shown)


packages/build-tools/src/buildErrors/userErrorHandlers.ts
```diff
@@ -112,33 +112,15 @@ export const userErrorHandlers: ErrorHandler<UserFacingError>[] = [
...
+      return new UserFacingError(
+        'EAS_BUILD_INCOMPATIBLE_PODS_ERROR',
+        `Compatible versions of some pods could not be resolved.
 You are seeing this error because either:
-  - Versions in the Podfile.lock cached by EAS do not match required values in Podspecs of some of the libraries. To fix that ${
-    usingDefaultCacheConfig
-      ? 'add the "cache.key" field (it can be set to any value) in the build profile in eas.json to invalidate the cache.'
-      : 'update value of the "cache.key" field in the build profile in eas.json to invalidate the cache.'
-  }
+  - Versions in the Podfile.lock cached by EAS do not match required values in Podspecs of some of the libraries. To fix that you can re-run build command with "--clear-cache" option, or select "Clear cache and retry build" on the build page.
```
wkozyra95, line 120: I'm not sure if I should mention that restart via website will work only for some time after build.

szdziedzic (approved):

packages/build-tools/src/buildErrors/userErrorHandlers.ts
brentvatne, line 120:
```suggestion
  - Versions in the Podfile.lock cached by EAS do not match required values in Podspecs of some installed libraries. To fix this, you can re-run build command with "--clear-cache" option, or select "Clear cache and retry build" on the build page.
```

brentvatne: i think it should be fine, but it looks like we in a follow up we should update the "Retry build" and "Clear cache & retry build" buttons to be disabled show some information about why they are disabled when the retry window has closed.

https://linear.app/expo/issue/ENG-6446/update-build-retry-buttons-to-be-visible-but-disabled-when-not

brentvatne (approved):

packages/build-tools/src/buildErrors/userErrorHandlers.ts
```diff
@@ -112,33 +112,15 @@ export const userErrorHandlers: ErrorHandler<UserFacingError>[] = [
...
+      return new UserFacingError(
+        'EAS_BUILD_INCOMPATIBLE_PODS_ERROR',
+        `Compatible versions of some pods could not be resolved.
 You are seeing this error because either:
-  - Versions in the Podfile.lock cached by EAS do not match required values in Podspecs of some of the libraries. To fix that ${
-    usingDefaultCacheConfig
-      ? 'add the "cache.key" field (it can be set to any value) in the build profile in eas.json to invalidate the cache.'
-      : 'update value of the "cache.key" field in the build profile in eas.json to invalidate the cache.'
-  }
+  - Versions in the Podfile.lock cached by EAS do not match required values in Podspecs of some installed libraries. To fix this, you can re-run build command with "--clear-cache" option, or select "Clear cache and retry build" on the build page.
```
brentvatne, line 120:
```suggestion
  - Versions in the cached Podfile.lock do not match required values in Podspecs of some installed libraries. To fix this, you can re-run build command with "--clear-cache" option, or select "Clear cache and retry build" on the build page.
```

brentvatne: small wording tweak

brentvatne (approved):

PR #152, opened 2022-09-26
[cocoapods-nexus-plugin] blacklist TrezorCrypto pod

szdziedzic:
# Why

TrezorCrypto pod doesn't work with our proxy and I've seen some errors about it in Sentry.

# How

Blacklist TrezorCrypto pod

# Test Plan

Tested locally


wkozyra95 (approved):
Based on the errors you have seen, can we write sth that is matching cocoapods errors without false positives?

szdziedzic:
@wkozyra95 I think it might be impossible to get rid of all false positives because logs for runs that fail because of our proxy vs because of user error looks the same and produce the same error. I need to manually validate it against CDN to make sure it is an error caused by our proxy.

PR #211, opened 2023-03-31
[build-tools] Fix timeout on doctor command

wkozyra95:
# Why

Fixes for doctor command

# How

- When process is timing out current implementation does not kill it properly, most likely it's sth npx is doing. Even SIGKILL does not work, it only kills npx process, but actual code keeps running. Given that I'm looking for all children of a process recursively and killing them all.
- New doctor does not support sdks <46, so I'm falling back to the old doctor
- Number of timeouts increased very significantly, plus we are using npx now, so there is additional overhead of running a doctor command, so I'm increasing timeout from 20 seconds to 30.

# Test Plan

Tested with local builds for both new sdks and old ones (with timeout small enough to trigger and when command had time to finish)
Did not test on macOS

### DIFF
--- a/packages/build-tools/src/common/setup.ts
+++ b/packages/build-tools/src/common/setup.ts
@@ -1,19 +1,21 @@
-import spawn, { SpawnResult } from '@expo/turtle-spawn';
+import spawn, { SpawnPromise, SpawnResult } from '@expo/turtle-spawn';
 import { BuildPhase, Ios, Job, Platform } from '@expo/eas-build-job';
 import { BuildTrigger } from '@expo/eas-build-job/dist/common';
+import nullthrows from 'nullthrows';
 
 import { BuildContext } from '../context';
 import { deleteXcodeEnvLocalIfExistsAsync } from '../ios/xcodeEnv';
 import { Hook, runHookIfPresent } from '../utils/hooks';
 import { createNpmrcIfNotExistsAsync, logIfNpmrcExistsAsync } from '../utils/npmrc';
 import { isAtLeastNpm7Async } from '../utils/packageManager';
-import { readPackageJson } from '../utils/project';
+import { readPackageJson, shouldUseGlobalExpoCli } from '../utils/project';
+import { getAllChildrenRecursiveAsync } from '../utils/processes';
 
 import { prepareProjectSourcesAsync } from './projectSources';
 import { installDependenciesAsync } from './installDependencies';
 import { configureEnvFromBuildProfileAsync, runEasBuildInternalAsync } from './easBuildInternal';
 
-const MAX_EXPO_DOCTOR_TIMEOUT_MS = 20 * 1000;
+const MAX_EXPO_DOCTOR_TIMEOUT_MS = 30 * 1000;
 
 class DoctorTimeoutError extends Error {}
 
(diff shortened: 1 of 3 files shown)


szdziedzic (approved):

packages/build-tools/src/utils/processes.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export async function getAllChildrenRecursiveAsync(ppid: number): Promise<number[]> {
+  const children: number[] = [ppid];
+  let shouldRetry = true;
+  while (shouldRetry) {
+    const pids = await getChildrenPidsAsync(children);
+    shouldRetry = false;
+    for (const pid of pids) {
+      if (!children.includes(pid)) {
```
dsokal, line 21: You can use a set for `children`.

packages/build-tools/src/utils/processes.ts
```diff
@@ -0,0 +1,28 @@
...
+    stdio: 'pipe',
+  });
+  return result.stdout
+    .toString()
+    .split('\n')
+    .map((i) => Number(i.trim()))
+    .filter((i) => i);
+}
+
+export async function getAllChildrenRecursiveAsync(ppid: number): Promise<number[]> {
```
dsokal, line 14: This function is not recursive, so why the suffix?

packages/build-tools/src/utils/processes.ts
```diff
@@ -0,0 +1,28 @@
...
+  while (shouldRetry) {
+    const pids = await getChildrenPidsAsync(children);
+    shouldRetry = false;
+    for (const pid of pids) {
+      if (!children.includes(pid)) {
+        shouldRetry = true;
+        children.push(pid);
+      }
+    }
+  }
```
dsokal, line 26: Can you use the actual recursion to make this more readable?

The problem I can see in this implementation (if I'm not missing anything) is that you repeatedly call `getChildrenPidsAsync` for the same pids. You iterate the `children` list and I believe the result for the same pid can not change.

packages/build-tools/src/common/setup.ts
```diff
@@ -85,16 +87,33 @@ async function runExpoDoctor<TJob extends Job>(ctx: BuildContext<TJob>): Promise
...
+          env: ctx.env,
+        },
+        isAtLeastNpm7
+      );
+    }
+    timeout = setTimeout(async () => {
       timedOut = true;
-      promise.child.kill();
+      const ppid = nullthrows(promise.child.pid);
+      (await getAllChildrenRecursiveAsync(ppid)).forEach((pid) => {
```
dsokal, line 114: IMO the name is misleading. I wouldn't expect this function to return the parent process pid.
- If you want the function to return the parent pid as well, name it `getParentAndChildrenPidsAsync`.
- If you want to create the list with parent pid and the result of the function here, name it `getChildrenPidsAsync`.

packages/build-tools/src/utils/processes.ts
```diff
@@ -0,0 +1,28 @@
...
+
+async function getChildrenPidsAsync(parentPids: number[]): Promise<number[]> {
+  const result = await spawn('pgrep', ['-P', parentPids.join(',')], {
+    stdio: 'pipe',
+  });
+  return result.stdout
+    .toString()
+    .split('\n')
+    .map((i) => Number(i.trim()))
+    .filter((i) => i);
```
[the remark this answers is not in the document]
wkozyra95: there is no such thing as pid 0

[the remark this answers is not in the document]
dsokal: Yes, there is https://www.google.com/search?q=pid+0&oq=pid+0 :D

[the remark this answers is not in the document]
wkozyra95: in kernel internals, not in userspace

PR #231, opened 2023-06-05
Replace `@expo/bunyan` with `bunyan`

sjchmiela:
# Why

Use maintained dependencies.

# How

Inspected differences between `@expo/bunyan` and `bunyan`. There are two:

- `@expo/bunyan` adds unique `id` to every log ([commit](https://github.com/trentm/node-bunyan/commit/e061d981f14d19376d99bd4e4bd7404dab553132)). Do we know if we depend on `id` being present anywhere, or why was it added? I tried to find any usage in `universe`, but didn't find any.
- [removes](https://github.com/trentm/node-bunyan/commit/d8a22b369a98fb70a2ed5b48b4548216e4e62bd8)/[disables](https://github.com/trentm/node-bunyan/commit/428e78233d2049f780c2475a51c392b67704f08c) dtrace — if I understand correctly, Dominik hinted in the issue description that it could have been removed because `expo-cli` depended on `@expo/bunyan`. My wishful guess is that this is no longer an issue, but it's just a guess…

What do you think? We [_could_](https://github.com/expo/eas-build/pull/231/commits/c5b7a8d5f89a56eab3f2799c30ba97a6e2d6c65d) keep adding unique `id` using private `bunyan` API (but hence it's private I don't think it's a good idea).

# Test Plan

Built a test Expo app locally with [the script suggested in the docs](https://github.com/expo/eas-build/blob/df67875b24a643ba821d1092a40683107c518391/DEVELOPMENT.md#L14-L22).

### DIFF
--- a/packages/build-tools/src/__mocks__/@expo/logger.ts
+++ b/packages/build-tools/src/__mocks__/@expo/logger.ts
@@ -1,4 +1,4 @@
-import bunyan from '@expo/bunyan';
+import bunyan from 'bunyan';
 
 export function createLogger(): bunyan {
   const logger = {
--- a/packages/local-build-plugin/package.json
+++ b/packages/local-build-plugin/package.json
@@ -25,10 +25,10 @@
   "license": "BUSL-1.1",
   "dependencies": {
     "@expo/build-tools": "1.0.18",
-    "@expo/bunyan": "^4.0.0",
     "@expo/eas-build-job": "1.0.13",
     "@expo/spawn-async": "^1.7.0",
     "@expo/turtle-spawn": "1.0.13",
+    "bunyan": "^1.8.15",
     "chalk": "^4.1.2",
     "env-paths": "^2.2.1",
     "fs-extra": "^11.1.0",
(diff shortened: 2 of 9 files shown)


wkozyra95 (approved):
LGTM, but this change affects all the services, not only the logger, so you will need to test it e2e eventually, local build is not enough.

With that said, I'm not saying that I expect it tested before merging. You can do that after publishing those packages, or even directly on turtle-v2 staging.

Just for the context how you could test it:
- before merging this PR
  - `yarn link` in each package (you will need to do it only once)
  - run https://github.com/expo/turtle-v2/blob/main/scripts/link-all-eas-dependencies.sh
  - start local turtle and www (let me know if you have trouble setting it up, it's quite involved to setup everything)
- after publishing this PR
  - update packages in turtle-v2 repo using https://github.com/expo/turtle-v2/blob/main/scripts/update-all-eas-dependencies.sh or wait for PR with updated dependencies that will be create automatically.
  - test locally or deploy from a branch by commenting out conditions in circle config file
- merge to main, wait for all services to get deployed to staging and test there.

PR #237, opened 2023-06-19
[steps] refactor context to simplify injecting logic into stpes library

wkozyra95:
# Why

Refactor context in steps library to make it easier to inject logic from build-tools.

# How

- Separate BuildStepContext into 2 separate types (BuildStepContext - per step context, BuildGlobalContext- context global for entire build)
- Move most BuildStepContext constructor arguments into BuildContextProvider
- Stop using buildId in steps library. As far as I can see that value was only used as random id, so replacing it with just uuidv4() should be fine.
- Provide a way to override env variables in the middle of the build.

# Test Plan

- tests
- cli.sh with some of examples
- run local turtle and running custom build from turtle-v2-example

### DIFF
--- a/packages/steps/src/BuildStepContext.ts
+++ b/packages/steps/src/BuildStepContext.ts
@@ -2,30 +2,53 @@ import os from 'os';
 import path from 'path';
 
 import { bunyan } from '@expo/logger';
+import { v4 as uuidv4 } from 'uuid';
 
 import { BuildStep } from './BuildStep.js';
 import { parseOutputPath } from './utils/template.js';
 import { BuildStepRuntimeError } from './errors.js';
 import { BuildRuntimePlatform } from './BuildRuntimePlatform.js';
+import { BuildStepEnv } from './BuildStepEnv.js';
 
-export class BuildStepContext {
+export interface ExternalBuildContextProvider {
+  readonly projectSourceDirectory: string;
+  readonly projectTargetDirectory: string;
+  readonly defaultWorkingDirectory: string;
+  readonly runtimePlatform: BuildRuntimePlatform;
+  readonly logger: bunyan;
+
+  readonly staticContext: any;
+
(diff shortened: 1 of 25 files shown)


packages/steps/src/BuildStepContext.ts
```diff
@@ -2,30 +2,53 @@ import os from 'os';
...
+  readonly runtimePlatform: BuildRuntimePlatform;
+  readonly logger: bunyan;
+
+  readonly staticContext: any;
+
+  readonly env: BuildStepEnv;
+  updateEnv(env: BuildStepEnv): void;
+}
+
+export class BuildGlobalContext {
```
wkozyra95, line 26: I'm open to name suggestions. I considered also
- BuildStepGlobalContext
- BuildContext (it's no really clear when we use it in build-tools which build context is which)

packages/steps/src/BuildStepContext.ts
```diff
@@ -2,30 +2,53 @@ import os from 'os';
...
+import { v4 as uuidv4 } from 'uuid';
 
 import { BuildStep } from './BuildStep.js';
 import { parseOutputPath } from './utils/template.js';
 import { BuildStepRuntimeError } from './errors.js';
 import { BuildRuntimePlatform } from './BuildRuntimePlatform.js';
+import { BuildStepEnv } from './BuildStepEnv.js';
 
-export class BuildStepContext {
+export interface BuildContextProvider {
```
wkozyra95, line 13: Not really sure about the name here too. Context provider suggests that it will provide a context, but it's in fact providing values that context needs internally.

I could drop a provider here, and name it sth like e.g. ExternalBuildContext or maybe BuildContextImpl or sth like that, but this would mean we 3 separate context objects here, which is a bit messy.

szdziedzic (approved):

PR #239, opened 2023-06-19
[ENG-8848][build-tools] implement `eas/build_react_native_app` for iOS without credentials

szdziedzic:
# Why

https://linear.app/expo/issue/ENG-8848/implement-easbuild-react-native-app-function

# How

Reuse existing functions from `@expo/build-tools` to build iOS app in `eas/build_react_native_app` function (without credentials)

# Test Plan

Run build locally


packages/build-tools/src/steps/utils/ios/fastlane.ts
```diff
@@ -0,0 +1,81 @@
+import path from 'path';
```
wkozyra95, line 1: I'm not convinced this is a good dir structure, but I think the best approach is to move stuff around when we have full build flow implemented and leave if for now how it is.

wkozyra95 (approved):

PR #242, opened 2023-06-21
[ENG-9019][steps] add support for step envs

szdziedzic:
# Why

https://linear.app/expo/issue/ENG-9019/add-support-for-envs-in-step-config

# How

Add support for step envs

This step
```yaml
- run:
    id: id123
    command: echo "$FOO $BAR"
    env:
       FOO: bar
       BAR: baz
```
will print `bar baz`

You can also pass envs to function calls
```yaml
- say_hi:
    env:
      ENV1: value1
      ENV2: value2
    inputs:
      name: Dominik
```
# Test Plan

Automated tests

### DIFF
--- a/packages/steps/src/BuildStep.ts
+++ b/packages/steps/src/BuildStep.ts
@@ -63,6 +63,7 @@ export class BuildStep {
   public readonly fn?: BuildStepFunction;
   public readonly shell: string;
   public readonly ctx: BuildStepContext;
+  public readonly env: Record<string, string>;
   public status: BuildStepStatus;
 
   private readonly internalId: string;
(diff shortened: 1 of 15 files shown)


sjchmiela (commented):
Looks reasonable to me!

sjchmiela (approved):
Whoops, meant to approve

packages/steps/src/BuildStep.ts
```diff
@@ -63,6 +63,7 @@ export class BuildStep {
   public readonly fn?: BuildStepFunction;
   public readonly shell: string;
   public readonly ctx: BuildStepContext;
+  public readonly env: Record<string, string>;
```
wkozyra95, line 66: we already have BuildStepEnv type

wkozyra95 (approved):

PR #245, opened 2023-06-22
[build-tools] Support set-env in regular builds

wkozyra95:
# Why

Add set-env support for the build.

# How

Similar script as implementation for custom builds.
After every build phase check directory with envs, add them to the context, and delete files from the directory.

# Test Plan

Started local turtle and www, added `set-env` call in pre install hook, added `env` call in post install hook.


packages/build-tools/src/context.ts
```diff
@@ -301,6 +324,29 @@ export class BuildContext<TJob extends Job> {
     this.buildPhaseHasWarnings = false;
   }
 
+  private async collectEnvVariablesAsync(): Promise<void> {
```
szdziedzic, line 327:
```suggestion
  private async collectAndUpdateEnvVariablesAsync(): Promise<void> {
```

szdziedzic (approved):
LGTM, one small naming suggestion:

PR #253, opened 2023-07-27
[ENG-9477][build-tools] implement `eas/configure-eas-update-if-installed` function

szdziedzic:
# Why

https://linear.app/expo/issue/ENG-9477/implement-easconfigure-expo-updates-if-installed

The next step towards making custom builds public

# How

Reimplement [currently existing logic](https://github.com/expo/eas-build/blob/34a9ba17497281793d3fa1f31808911b666dadb6/packages/build-tools/src/utils/expoUpdates.ts#L140-L187) to configure EAS update during the build process

Drop support for classic updates in the new implementation

# Test Plan

Add tests

### DIFF
--- a/packages/build-tools/src/steps/easFunctions.ts
+++ b/packages/build-tools/src/steps/easFunctions.ts
@@ -12,6 +12,7 @@ import { createRunGradleBuildFunction } from './functions/utils/runGradle';
 import { createPrebuildBuildFunction } from './functions/eas/prebuild';
 import { createBuildReactNativeAppBuildFunction } from './functions/eas/buildReactNativeApp';
 import { createFindAndUploadBuildArtifactsBuildFunction } from './functions/eas/findAndUploadBuildArtifacts';
+import { configureExpoUpdatesIfInstalledFunction } from './functions/eas/configureExpoUpdatesIfInstalled';
 
 export function getEasFunctions(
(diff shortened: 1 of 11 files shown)


sjchmiela (approved):
As far as I understand, it just copies part of `build-tools/utils/` related to EAS Update into `steps/utils`?

Would it make sense to try to remove the old copy and use the new copy instead?

szdziedzic:
>Would it make sense to try to remove the old copy and use the new copy instead?

The old copy is still used by the "standard" build process. For custom builds, we are just reimplementing these utilities with the intent of making a "standard" build just an implicit custom build (using predefined build configs) at some point in time. Once this is done we will basically remove all the code used by the old build process.

We decided to go with the reimplementation to not cause any issues when trying to make the same code interoperative between the new build process and the old build process. We also drop support for legacy stuff like custom updates when reimplementing functionalities (what we couldn't do when making it interoperative).

szdziedzic:
>As far as I understand, it just copies part of build-tools/utils/ related to EAS Update into steps/utils?

I copied most of the code, but also dropped support for classic updates and made it independent from `BuildContext` used by the standard build process (which isn't available in the custom builds, at least not in the user-defined functions, so we don't want to use it whenever we can).

PR #274, opened 2023-08-31
[ENG-9955][eas-build-job] add build phase to external build error

szdziedzic:
# Why

https://linear.app/expo/issue/ENG-9955/link-to-build-logs-from-eas-cli-when-build-fails-for-unknown-reason

# How

Add information about the build phase to the `ExternalBuildError` interface. Add mappings from build phase enum to build phase display names and ids so they can be shared between multiple repos using `eas-build-job` package

# Test Plan

Tests

### DIFF
--- a/packages/eas-build-job/src/errors.ts
+++ b/packages/eas-build-job/src/errors.ts
@@ -13,6 +13,7 @@ export interface ExternalBuildError {
   errorCode: string;
   message: string;
   docsUrl?: string;
+  buildPhase?: BuildPhase;
 }
 
 interface BuildErrorDetails {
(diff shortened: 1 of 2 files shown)


packages/eas-build-job/src/errors.ts
```diff
@@ -57,8 +59,13 @@ export class UserFacingError extends Error {
 }
 
 export class UnknownError extends UserFacingError {
-  constructor() {
-    super(ErrorCode.UNKNOWN_ERROR, 'Unknown error. See logs for more information.');
+  constructor(buildPhase?: BuildPhase) {
+    super(
+      ErrorCode.UNKNOWN_ERROR,
+      buildPhase
```
radoslawkrzemien, line 65: Maybe also check if the `buildPhase` is in the keys of `buildPhaseDisplayName`, just to cover the case when `buildPhase` is set but the mapping is missing

radoslawkrzemien (approved):

PR #278, opened 2023-09-12
[ENG-10069] add bun option to eas.json

kadikraman:
# Why

Bun is fast, they hype is real and people want to use it! https://bun.sh/

# How

Mirrored the implementation for `pnpm` version https://github.com/expo/eas-build/pull/254


szdziedzic (commented):
Nice! I think that you also need to update our [`Joi`](https://github.com/hapijs/joi) schema for both iOS and Android to make sure that we validate that `bun` is an optional `string` value 😄 

https://github.com/expo/eas-build/blob/main/packages/eas-build-job/src/ios.ts#L61C1-L61C1

PR #281, opened 2023-09-27
[eas-build-job] [ENG-10225] add requiredPackageManager to build metadata

kadikraman:
# Why

We want to know what the user's package manager is going to be at the "install custom tools" step of the build process, so we're determining it based on the user's lockfiles and adding it to build metadata.

1. 👉  `eas-build-job` https://github.com/expo/eas-build/pull/281
2. `www` https://github.com/expo/universe/pull/13522
3. `eas-cli` https://github.com/expo/eas-cli/pull/2067

# How

Added the field.

# Test Plan

Unit tests and building locally.

### DIFF
--- a/packages/eas-build-job/src/metadata.ts
+++ b/packages/eas-build-job/src/metadata.ts
@@ -149,6 +149,11 @@ export type Metadata = {
    * Indicates whether this is (likely, we can't be 100% sure) development client build.
    */
   developmentClient?: boolean;
+
+  /**
+   * Which package manager will be used for the build. Determined based on lockfiles in the project directory.
+   */
+  requiredPackageManager?: 'npm' | 'pnpm' | 'yarn' | 'bun';
 };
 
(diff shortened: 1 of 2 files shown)


packages/eas-build-job/src/metadata.ts
```diff
@@ -149,6 +149,11 @@ export type Metadata = {
    * Indicates whether this is (likely, we can't be 100% sure) development client build.
    */
   developmentClient?: boolean;
+
+  /**
+   * Which package manager will be used for the build. Determined based on lockfiles in the project directory.
+   */
+  requiredPackageManager?: 'npm' | 'pnpm' | 'yarn' | 'bun';
```
kadikraman, line 156: I decided to call it `requiredPackageManager` because this is the package manager that must get installed at the "install custom tools" step if not already, and if it isn't, we should fail the build.

PR #282, opened 2023-09-28
Add `NVM_NODEJS_ORG_MIRROR` which will be part of `turtle-v2`

sjchmiela:
# Why

Counterpart to https://github.com/expo/turtle-v2/pull/1448.

# How

Add `NVM_NODEJS_ORG_MIRROR` which we'll soon use in the real Turtle.

# Test Plan

None.

### DIFF
--- a/packages/local-build-plugin/src/build.ts
+++ b/packages/local-build-plugin/src/build.ts
@@ -25,8 +25,8 @@ export async function buildAsync(job: Job, metadata: Metadata): Promise<void> {
     const unfilteredEnv: Record<string, string | undefined> = {
       ...process.env,
       ...job.builderEnvironment?.env,
-      // We don't use Node.js mirror locally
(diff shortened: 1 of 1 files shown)


packages/local-build-plugin/src/build.ts
```diff
@@ -25,6 +25,8 @@ export async function buildAsync(job: Job, metadata: Metadata): Promise<void> {
     const unfilteredEnv: Record<string, string | undefined> = {
       ...process.env,
       ...job.builderEnvironment?.env,
+      // We don't use Node.js mirror locally
+      NVM_NODEJS_ORG_MIRROR: undefined,
```
sjchmiela, line 29:
```suggestion
      // We don't use Node.js mirror locally, but a user might
      NVM_NODEJS_ORG_MIRROR: process.env.NVM_NODEJS_ORG_MIRROR,
```

szdziedzic (approved):

PR #289, opened 2023-11-14
[ENG-9500] Implement save-cache and restore-cache functions

khamilowicz:
# Why

[ENG-9500](https://linear.app/expo/issue/ENG-9500/caching-for-custom-builds)

User should be able to save and restore files in their custom builds as steps.

This is part 1 of this feature. It provides two new custom build functions: `eas/save-cache` and `eas/restore-cache`. 

```
build:
  name: Foobar
  steps:
    - eas/checkout
    - eas/restore-cache:
      inputs:
        key: cache-key
        paths: node_modules
    - eas/install_node_module
    - eas/save-cache:
        inputs:
          key: cache-key
          paths: node_modules
```

Future PR will allow passing lists of files to `paths` parameter and define dynamic cache-key.

# How

This PR implements `eas/reastore-cache` and `eas/save-cache` functions and lays foundations for future PRs. 
`Cache` type is enhanced with `downloadUrls` attribute which will store mappings between cache keys and signed urls with cached files. Field will be populated by `launcher`. 
`CacheManager` now accepts two parameters: build context (which is either `BuildStepContext` or `BuildContext<Job>`) and `Cache`. Implementation of `CacheManager` using dynamic cache urls will be provided in future PRs. Until this happens, for compatibility, build context is using `any` type. 

# Test Plan

These changes are not meant to provide working functionality, but to provide foundation for new `GCSCacheManager` implementation.


packages/build-tools/src/steps/functions/cache.ts
```diff
@@ -0,0 +1,122 @@
...
+        id: 'paths',
+        required: true,
+        allowedValueTypeName: BuildStepInputValueTypeName.JSON,
+      }),
+    ],
+    fn: async (stepsCtx, { inputs }) => {
+      const cacheManager = stepsCtx.global.cacheManager;
+
+      if (!cacheManager) {
+        return;
```
sjchmiela, line 96: Should we log that something went wrong?

packages/steps/src/cacheUtils.ts
```diff
@@ -0,0 +1,31 @@
+import { bunyan } from '@expo/logger';
+
+export interface Cache {
+  disabled: boolean;
+  clear: boolean;
```
sjchmiela, line 5: What do these do?

packages/steps/src/cacheUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+  key?: string;
+  cacheDefaultPaths?: boolean;
+  customPaths?: string[];
+  paths: string[];
+  downloadUrls?: Record<string, string>;
```
sjchmiela, line 10:
```suggestion
```

packages/steps/src/cacheUtils.ts
```diff
@@ -0,0 +1,31 @@
+import { bunyan } from '@expo/logger';
+
+export interface Cache {
```
sjchmiela, line 3: Do we need to have two `Cache`s, one here and another one in `eas-build-job`?

packages/build-tools/src/steps/functions/cache.ts
```diff
@@ -0,0 +1,122 @@
...
+  BuildStepOutput,
+} from '@expo/steps';
+
+function createCacheKey(cacheKey: string, paths: string[]): string {
+  const hash = crypto.createHash('sha256');
+  hash.update(paths.sort().join(''));
+
+  const pathsHash = hash.digest('hex');
+
+  return `${cacheKey}-${pathsHash}`;
```
sjchmiela, line 16: Should this be the case? That cache key includes paths hash? Or should the cache key be just the key that the user provided? How does Github Actions caching work?

packages/steps/src/cacheUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+  logger: bunyan;
+  global: CacheableContextCtx;
+  workingdir: string;
+}
+
+export interface CacheManager {
+  // TOOD: Change any to CacheableContext
+  saveCache(ctx: any, cache?: Cache): Promise<boolean | void>;
+  // TOOD: Change any to CacheableContext
+  restoreCache(ctx: any, cache: Cache): Promise<boolean | void>;
```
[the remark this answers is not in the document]
khamilowicz: `void` is here only to keep compatibility with `GCSCacheManager` in turtle (https://github.com/expo/turtle-v2/blob/9e6fd370c862869fec034d7e1be1074f7d3e3105/src/services/worker/src/CacheManager.ts#L18)

packages/steps/src/cacheUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+  generateUrls?: boolean;
```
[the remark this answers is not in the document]
khamilowicz: Yes, it will tell new GCSCacheManager whether to use pregenerated upload/download urls or to generate new ones.

packages/steps/src/cacheUtils.ts
khamilowicz: It is part of the transition process: in order to avoid circular dependencies `Cache` should be defined in `steps`, but to keep compatibility I kept it in both places. It will be consolidated after changes to turtle

packages/build-tools/src/steps/functions/cache.ts
khamilowicz: GIthub Actions only uses provided key, but the problem would be when user adds new paths to cache without changing key. I can't imagine use case when user doesn't want to automatically update cache file content.

packages/steps/src/cacheUtils.ts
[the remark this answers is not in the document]
sjchmiela: My suggestion to make development of this feature easier will be to do as follows:
- make all necessary changes in `eas-build`. They can break API. We shouldn't try to deploy `turtle-v2` with new cache changes (bumped dependencies), but without your adjusting changes (making `GCSCacheManager` conform to new `restoreCache`.
- submit a pull request with upgraded dependencies and updated `turtle-v2` code.

This is the flow I used when developing generic artifacts:
- made a couple breaking changes to `eas-build` (https://github.com/expo/eas-build/pull/306, https://github.com/expo/eas-build/pull/309, https://github.com/expo/eas-build/pull/307)
- updated dependencies and adjusted `turtle-v2` in one PR (https://github.com/expo/turtle-v2/pull/1609).

This removes the complexity of trying to maintain two interfaces at the same time. What do you think, do you foresee any problems?

packages/build-tools/src/steps/functions/cache.ts
sjchmiela: Yea, makes sense. Also, GitHub _does_ take paths into consideration, see https://github.com/actions/cache?tab=readme-ov-file#cache-version. Maybe there's something else we should take in from what they do.

packages/build-tools/src/customBuildContext.ts
```diff
@@ -56,6 +59,9 @@ export class CustomBuildContext implements ExternalBuildContextProvider {
     this.runtimeApi = {
       uploadArtifacts: (...args) => buildCtx['uploadArtifacts'](...args),
     };
+    this.cacheManager = buildCtx.cacheManager;
```
sjchmiela, line 62: We could consider rolling `cacheManager` into `runtimeApi` for simplicity. Like, how many functions may we put into `runtimeApi`? I don't think the list will be too long.

khamilowicz:
@sjchmiela

packages/build-tools/src/steps/functions/cache.ts
```diff
@@ -0,0 +1,104 @@
...
+      BuildStepInput.createProvider({
+        id: 'paths',
+        required: true,
+        defaultValue: [],
+        allowedValueTypeName: BuildStepInputValueTypeName.JSON,
+      }),
+    ],
+    outputProviders: [
+      BuildStepOutput.createProvider({
+        id: 'cache_key',
```
sjchmiela, line 31: nit: I think `key` should be sufficient, same for `paths`.

Or, we could not add it because however users are providing us `key` and `paths`, they should be able to use the same technique anywhere else? This is not a strong opinion.

packages/build-tools/src/steps/functions/cache.ts
```diff
@@ -0,0 +1,104 @@
...
+      if (paths.length > 0) {
+        outputs.cache_paths.set(JSON.stringify(paths));
+      }
+
+      if (!cacheManager) {
+        stepsCtx.logger.warn('Cache manager is not available, skipping...');
+        return;
+      }
+
+      const job = stepsCtx.global.staticContext.job as BuildJob;
```
sjchmiela, line 54: If the function only makes sense for a `BuildJob` let's define argument to this function like https://github.com/expo/eas-build/blob/e15bbd61ccd025b2846d480e0cf454db962b934d/packages/build-tools/src/steps/functions/resolveBuildConfig.ts#L9-L11 so types require a `BuildJob` and we don't need to cast.

However, I think the bigger thing to update here is that I think this doesn't take into consideration updates to https://github.com/expo/turtle-v2/pull/1715, e.g. that cache download URLs are to be requested dynamically from the launcher?

packages/steps/build.sh
```diff
@@ -11,6 +11,7 @@ if [[ "$npm_lifecycle_event" == "prepack" ]]; then
   echo 'Removing "dist_commonjs" and "dist_esm" folders...'
   rm -rf dist_commonjs dist_esm
 fi
+rm -rf dist_commonjs dist_esm
```
sjchmiela, line 14: Why did you need to add this? Can we remove it or submit a different pull request explaining the reason for this addition?

packages/steps/src/__tests__/BuildConfig-test.ts
```diff
@@ -33,7 +33,7 @@ describe(readAndValidateBuildConfigAsync, () => {
     const config = await readAndValidateBuildConfigAsync(
       path.join(__dirname, './fixtures/build.yml'),
       {
-        externalFunctionIds: [],
+        externalFunctionIds: ['eas/save-cache'],
```
sjchmiela, line 36: I can see `eas/*-cache` functions being added as external functions in at least a couple of tests. Why wouldn't they be added as EAS Functions (in `steps/easFunctions.ts`)?

packages/steps/package.json
```diff
@@ -32,6 +32,7 @@
   "license": "BUSL-1.1",
   "devDependencies": {
     "@jest/globals": "^29.7.0",
+    "@types/glob": "^8.1.0",
```
sjchmiela, line 35: I think we're using `fast-glob` and this is not used/necessary?

packages/steps/src/inputFunctions.ts
```diff
@@ -0,0 +1,49 @@
...
+      );
+    })
+  );
+
+  return hash.digest('hex');
+}
+
+export const hashFiles = hashFilesAsync;
+
+export default async function callInputFunctionAsync(
```
sjchmiela, line 38: This (and changes in `template.ts`) are very cool, good job figuring this out! However:

1. Can we not include it in this pull request already? v0 of this pull request may only allow static strings and cache keys and that's ok.
2. Instead of this manual expression parsing and function calling I think we should try to use `jsepEval` on expressions. [This](https://github.com/expo/eas-build/blob/6376bdc493319be4d43e62d1b751fd2161e73cf4/packages/steps/src/BuildStep.ts#L297-L315) is how we're providing functions to `ifCondition`s. It should be simpler and more consistent with other places this way.

packages/steps/src/cli/cli.ts
```diff
@@ -23,7 +23,9 @@ export class CliContextProvider implements ExternalBuildContextProvider {
     public readonly projectSourceDirectory: string,
     public readonly projectTargetDirectory: string,
     public readonly defaultWorkingDirectory: string,
-    public readonly buildLogsDirectory: string
+    public readonly buildLogsDirectory: string,
+    public readonly buildDirectory: string,
+    public readonly projectRootDirectory: string
```
sjchmiela, line 28: Oh no, not more directories! haha

What is the reason to add them? Could we use one of the existing directories instead?

If not, can you add comments to those directories explaining what they are? The way we manage directories now is a mess and I find it hard to keep track of what is what.

packages/steps/src/cacheUtils.ts
```diff
@@ -0,0 +1,15 @@
...
+  clear: boolean;
+  key?: string;
+  cacheDefaultPaths?: boolean;
+  customPaths?: string[];
+  paths: string[];
+}
+
+export interface DynamicCacheManager {
+  saveCache(ctx: BuildStepContext, cache: Cache): Promise<boolean>;
+  restoreCache(ctx: BuildStepContext, cache: Cache): Promise<boolean>;
```
sjchmiela, line 14: What does `boolean` mean in this context? Could we use `<void>` and throw-and-catch on error?

packages/steps/src/cacheUtils.ts
sjchmiela: Is this still true after recent changes to Turtle? Why would we get a circular dependency if we used `Cache` from `eas-build-job`?

packages/steps/src/cacheUtils.ts
sjchmiela: Unresolving discussion. It remains unclear what does `disabled` or `clear` mean in this context.

packages/steps/src/__tests__/fixtures/build.yml
```diff
@@ -29,3 +29,8 @@ build:
         name: Use non-default shell
         shell: /nib/hsab
         command: echo 123
+    - eas/save-cache:
+        inputs:
+          key: cache-key
+          paths:
+            - src
```
sjchmiela, line 36: nit: new line at the end of file. (missing code editor setting on your side?)

```suggestion
            - src

```

packages/steps/src/__tests__/BuildWorkflowValidator-test.ts
```diff
@@ -519,4 +537,29 @@ describe(BuildWorkflowValidator, () => {
       `Custom function module path "/non/existent/module" for function "test" does not exist.`
     );
   });
+
+  test('non-existing  function module', async () => {
```
sjchmiela, line 541:
```suggestion
  test('non-existing function module', async () => {
```

packages/steps/src/BuildWorkflowValidator.ts
```diff
@@ -74,6 +75,28 @@ export class BuildWorkflowValidator {
...
+              }
+              if (fn && fn in inputFunctions) {
+                return null;
+              }
+              return new BuildConfigError(
+                `Input parameter "${currentStepInput.id}" for step "${
+                  currentStep.displayName
+                }" is set to "\${ ${fn}(${args
+                  .map((i) => `"${i}"`)
+                  .join(',')}) }" which is not a valid build-in function name.`
```
sjchmiela, line 94:
```suggestion
                  .join(',')}) }" which is not a valid built-in function name.`
```

packages/steps/src/BuildStepInput.ts
```diff
@@ -103,18 +108,65 @@ export class BuildStepInput<
     this.allowedValueTypeName = allowedValueTypeName;
   }
 
-  public get value(): BuildStepInputValueTypeWithRequired<T, R> {
+  public isFunctionCall(): boolean {
+    return this.rawValue?.toString().match(BUILD_STEP_FUNCTION_EXPRESSION_REGEXP) !== null;
```
sjchmiela, line 112: `rawValue?` means if there is no `rawValue` it will be `undefined` which will be `!== null` and will say it is a function call. Seems unexpected to me?

sjchmiela (changes requested):
Very cool, some questions and suggestions!

packages/steps/src/__tests__/BuildConfig-test.ts
khamilowicz: `readAndValidateBuildConfigAsync` doesn't know about eas functions, these are passed in `runCustomBuildAsync`(https://github.com/expo/eas-build/blob/afe570422d2c45a7478786e1aeb885a5014d90cf/packages/build-tools/src/builders/custom.ts#L39 and https://github.com/expo/eas-build/blob/afe570422d2c45a7478786e1aeb885a5014d90cf/packages/steps/src/BuildConfigParser.ts#L67).

sjchmiela: Hm, maybe let me rephrase my curiosity. Why do we only need to add `eas/save-cache` here and not all other EAS functions? Like, why this function is special and needs to be added here?

khamilowicz: Oh, that's because the file that we parse (`./fixtures/build.yml`) only contains steps using either `run` expressions or `eas/save-cache` function. Only functions that are used in template are checked for existence, and `run` is a special case (https://github.com/expo/eas-build/blob/afe570422d2c45a7478786e1aeb885a5014d90cf/packages/steps/src/BuildConfig.ts#L435)

sjchmiela: Aaaaah, it's a test file. Yea, sorry, didn't see that.

packages/build-tools/src/steps/functions/cache.ts
sjchmiela: - call to `restoreCache` depends on `job.cache` existence
- `restoreCache` receives `job.cache`

I thought dynamic caches are completely separate from `job.cache`.

PR #290, opened 2023-11-14
[ENG-10130][build-tools] use env vars from `.env` files when resolving app config same way expo CLI does

szdziedzic:
# Why

https://linear.app/expo/issue/ENG-10130/ensure-that-read-app-config-is-evaluated-in-a-way-that-it-loads-the

Currently, if the `.env` file is present in project files in the EAS Build and we have `VAL1` both present in the `.env` file and in project secrets, the value used during the app compilation, by prebuild command, and by the `npx expo config --type public` command will be the one from the `.env` file. However, our internal function to read app config won't see the value from the `.env` file and will use the value from secrets. 

It confuses our users because in the `Read app config` build phase they will see the app config resolved by us using our internal function which doesn't use `.env` files, but in some cases if the same envs are present as both secret and in `.env` file it might be different for what actually will be used during prebuild and compilation. These phases will see the envs in the `.env` file and use them when they clash with envs as secrets.

# How

Install `@expo/env` as a project dependency. Use the `load` function from `@expo/env` to load values from the `.env` file when resolving app config.

# Test Plan

Run builds locally and see which value will be used.

`VAL1` is present in both the project secret and in the `.env` file

`.env` file:
```
VAL1=notsosecret
```

project secrets:
```
VAL1=secretvalue (masked)
```

Before:
Screenshot 2023-11-14 at 17 02 17

After:
Screenshot 2023-11-14 at 17 02 58

It matches the output of the `npx expo config --type public` command.

### DIFF
--- a/packages/build-tools/src/utils/appConfig.ts
+++ b/packages/build-tools/src/utils/appConfig.ts
@@ -19,11 +19,12 @@ export function readAppConfig({
   const originalProcessCwd = process.cwd;
   const originalStdoutWrite = process.stdout.write;
   const originalStderrWrite = process.stderr.write;
+  const originalProcessEnv = process.env;
 
   const stdoutStore: { text: string; level: LoggerLevel }[] = [];
   const shouldLoadEnvVarsFromDotenvFile = sdkVersion && semver.satisfies(sdkVersion, '>=49');
   const envVarsFromDotenvFile = shouldLoadEnvVarsFromDotenvFile ? load(projectDir) : {};
-  const newEnvsToUse = { ...envVarsFromDotenvFile, ...env };
+  const newEnvsToUse = { ...env, ...envVarsFromDotenvFile };
   try {
     for (const [key, value] of Object.entries(newEnvsToUse)) {
       process.env[key] = value;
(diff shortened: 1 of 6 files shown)


packages/build-tools/src/utils/appConfig.ts
```diff
@@ -1,17 +1,31 @@
...
   const originalProcessCwd = process.cwd;
   const originalStdoutWrite = process.stdout.write;
   const originalStderrWrite = process.stderr.write;
 
   const stdoutStore: { text: string; level: LoggerLevel }[] = [];
+  const shouldLoadEnvVarsFromDotenvFile = sdkVersion && semver.satisfies(sdkVersion, '>=49');
+  const envVarsFromDotenvFile = shouldLoadEnvVarsFromDotenvFile ? load(projectDir) : {};
   try {
-    process.env = { ...env };
+    process.env = { ...env, ...envVarsFromDotenvFile };
```
byCedric, line 28: This might have unexpected side effects in Bun, preferably we don’t overwrite `process.env` with an object. Instead, setting the properties from the env file to the existing object might be better here.

See https://github.com/expo/expo/pull/25393 for more info on this “mindf*ck” 🙈

szdziedzic: @byCedric do you mean something like the changes I added in my last commit?

brentvatne (approved):

packages/build-tools/src/utils/appConfig.ts
byCedric: Yes, exactly 😄

byCedric (approved):

PR #291, opened 2023-11-16
[ENG-10477][steps][build-tools] use correct default working directory if project wasn't checked out

szdziedzic:
# Why

https://github.com/expo/eas-build/blob/1c3f2019ad05de2c1fc3dda666cc0dcafdd3e492/packages/build-tools/src/customBuildContext.ts#L52

We set the `defaultWorkingDirectory` in our external build context provider as the output of the `getReactNativeProjectDirectory()` function. This function will return `path.join(workingDir, 'build') + relativePathToEasJsonInYourGitRepo`. But unfortunately, if the user didn't call `checkout` this `relativePathToEasJsonInYourGitRepo` might not exist.

# How

Use a boolean value to monitor whether the project was checked out. If it was use the previous defaultWorkingDirectory. If not use project target dir (`workingDir + '/build'`) which always exists and it is the dir when all of the operations should be performed.

# Test Plan

Automated tests

### DIFF
--- a/packages/build-tools/src/steps/functions/checkout.ts
+++ b/packages/build-tools/src/steps/functions/checkout.ts
@@ -15,6 +15,7 @@ export function createCheckoutBuildFunction(): BuildFunction {
           overwrite: true,
         }
       );
+      stepsCtx.global.markAsCheckedOut();
     },
   });
 }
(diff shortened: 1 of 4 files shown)


sjchmiela (approved):
I agree it makes the most sense. I wish we only had to consider `.eas` in repository root and have `defaultWorkingDirectory` _always_ be `projectTargetDirectory`. Like, in my opinion it would be best if working directory was always root and the user did:

```
- eas/run_fastlane:
    working-directory: mobile
```

Maybe some other day.

PR #292, opened 2023-11-21
[ENG-10699][steps] introduce conditional checks for generic job steps

szdziedzic:
# Why

Please describe the motivation for this PR, and link to relevant GitHub issues, forums posts, or feature requests.

# How

How did you build this feature or fix this bug and why?

# Test Plan

Please describe how you tested this change and how a reviewer could reproduce your test, especially if this PR does not include automated tests! If possible, please also provide terminal output and/or screenshots demonstrating your test/reproduction.

### DIFF
--- a/packages/steps/src/BuildConfig.ts
+++ b/packages/steps/src/BuildConfig.ts
@@ -14,7 +14,10 @@ import {
   BuildStepInputValueType,
 } from './BuildStepInput.js';
 import { BuildStepEnv } from './BuildStepEnv.js';
-import { BUILD_STEP_OR_BUILD_GLOBAL_CONTEXT_REFERENCE_REGEX } from './utils/template.js';
+import {
+  BUILD_STEP_IF_CONDITION_EXPRESSION_REGEXP,
+  BUILD_STEP_OR_BUILD_GLOBAL_CONTEXT_REFERENCE_REGEX,
+} from './utils/template.js';
 
 export type BuildFunctions = Record<string, BuildFunctionConfig>;
 
@@ -55,6 +58,7 @@ export type BuildFunctionCallConfig = {
   workingDirectory?: string;
   shell?: string;
   env?: BuildStepEnv;
+  if?: string;
 };
 
 export type BuildStepInputs = Record<string, BuildStepInputValueTypeWithRequired>;
@@ -191,6 +195,7 @@ const BuildFunctionCallSchema = Joi.object({
   workingDirectory: Joi.string(),
   shell: Joi.string(),
   env: Joi.object().pattern(Joi.string(), Joi.string().allow('')),
+  if: Joi.string().pattern(BUILD_STEP_IF_CONDITION_EXPRESSION_REGEXP, 'step if condition regex'),
 }).rename('working_directory', 'workingDirectory');
 
 const BuildStepConfigSchema = Joi.any<BuildStepConfig>()
--- a/packages/steps/src/BuildStep.ts
+++ b/packages/steps/src/BuildStep.ts
@@ -21,7 +21,10 @@ import {
   saveScriptToTemporaryFileAsync,
 } from './BuildTemporaryFiles.js';
 import { spawnAsync } from './utils/shell/spawn.js';
-import { interpolateWithInputs } from './utils/template.js';
+import {
+  getSelectedStatusCheckFromIfStatementTemplate,
+  interpolateWithInputs,
+} from './utils/template.js';
 import { BuildStepRuntimeError } from './errors.js';
 import { BuildStepEnv } from './BuildStepEnv.js';
 import { BuildRuntimePlatform } from './BuildRuntimePlatform.js';
@@ -125,6 +128,7 @@ export class BuildStep extends BuildStepOutputAccessor {
   public readonly shell: string;
   public readonly ctx: BuildStepContext;
   public readonly env: BuildStepEnv;
+  public readonly ifCondition?: string;
   public status: BuildStepStatus;
 
   private readonly internalId: string;
@@ -177,6 +181,7 @@ export class BuildStep extends BuildStepOutputAccessor {
       shell,
       supportedRuntimePlatforms: maybeSupportedRuntimePlatforms,
       env,
+      ifCondition,
     }: {
       id: string;
       name?: string;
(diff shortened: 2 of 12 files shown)


packages/steps/src/utils/template.ts
```diff
@@ -24,6 +25,25 @@ export function interpolateWithOutputs(
   return interpolate(templateString, BUILD_STEP_OUTPUT_EXPRESSION_REGEXP, fn);
 }
 
+/**
+ * Returns the selected status check from the if statement template.
+ * Example: `${ success() }` -> `success`
+ */
+// TODO @szdziedzic: add support for cancelled
```
sjchmiela, line 32: Should we create a Linear task for it already? Maybe replace this with a link to Linear task? There's enough `// TODO`s as it is. 😅

packages/steps/src/utils/template.ts
```diff
@@ -9,6 +9,7 @@ export const BUILD_STEP_INPUT_EXPRESSION_REGEXP = /\${\s*(inputs\.[\S]+)\s*}/;
 export const BUILD_STEP_OUTPUT_EXPRESSION_REGEXP = /\${\s*(steps\.[\S]+)\s*}/;
 export const BUILD_GLOBAL_CONTEXT_EXPRESSION_REGEXP = /\${\s*(eas\.[\S]+)\s*}/;
 export const BUILD_STEP_OR_BUILD_GLOBAL_CONTEXT_REFERENCE_REGEX = /\${\s*((steps|eas)\.[\S]+)\s*}/;
+export const BUILD_STEP_IF_CONDITION_EXPRESSION_REGEXP = /\${\s*((always|success|failure)\(\))\s*}/;
```
sjchmiela, line 12: Do we need the encompassing capture group? Would 
```suggestion
export const BUILD_STEP_IF_CONDITION_EXPRESSION_REGEXP = /\${\s*(always|success|failure)\(\)\s*}/;
```
be sufficient?

packages/steps/src/BuildWorkflow.ts
```diff
@@ -16,8 +16,22 @@ export class BuildWorkflow {
   }
 
   public async executeAsync(): Promise<void> {
+    let maybeError: Error | null = null;
+    let hasAnyPreviousStepsFailed = false;
```
sjchmiela, line 20: nit
```suggestion
    let hasAnyPreviousStepFailed = false;
```

packages/steps/src/BuildWorkflow.ts
```diff
@@ -16,8 +16,22 @@ export class BuildWorkflow {
...
+          await step.executeAsync();
+        } catch (err: any) {
+          maybeError = err;
+          hasAnyPreviousStepsFailed = true;
+        }
+      } else {
+        step.skip();
+      }
+    }
+    if (maybeError) {
```
sjchmiela, line 33: nit: so it's tiny bit clearer it's outside of the `for` (had to check)
```suggestion

    if (maybeError) {
```

packages/steps/src/BuildWorkflow.ts
```diff
@@ -16,8 +16,22 @@ export class BuildWorkflow {
...
   public async executeAsync(): Promise<void> {
+    let maybeError: Error | null = null;
+    let hasAnyPreviousStepsFailed = false;
     for (const step of this.buildSteps) {
-      await step.executeAsync();
+      if (step.shouldExecuteStep(hasAnyPreviousStepsFailed)) {
+        try {
+          await step.executeAsync();
+        } catch (err: any) {
+          maybeError = err;
```
sjchmiela, line 26: Should we error with the first or last error? Why?

packages/steps/src/BuildStep.ts
```diff
@@ -281,6 +288,35 @@ export class BuildStep extends BuildStepOutputAccessor {
...
+    this.status = BuildStepStatus.SKIPPED;
+    this.ctx.logger.info(
+      { marker: BuildStepLogMarker.START_STEP },
+      'Executing build step "${this.displayName}"'
+    );
+    this.ctx.logger.info(`Skipped build step "${this.displayName}"`);
+    this.ctx.logger.info(
+      { marker: BuildStepLogMarker.END_STEP, result: BuildStepStatus.SKIPPED },
+      `Skipped build step "${this.displayName}"`
+    );
```
sjchmiela, line 317: We need those three logs to start and end step phase? My first reaction was "why not one `.info()`".

packages/steps/src/BuildConfig.ts
```diff
@@ -191,6 +195,7 @@ const BuildFunctionCallSchema = Joi.object({
   workingDirectory: Joi.string(),
   shell: Joi.string(),
   env: Joi.object().pattern(Joi.string(), Joi.string().allow('')),
+  if: Joi.string().pattern(BUILD_STEP_IF_CONDITION_EXPRESSION_REGEXP, 'step if condition regex'),
```
sjchmiela, line 198:
```suggestion
  if: Joi.string().pattern(BUILD_STEP_IF_CONDITION_EXPRESSION_REGEXP, 'step "if" condition regex'),
```
hm?

sjchmiela (approved):

packages/steps/src/utils/template.ts
szdziedzic: Added: https://linear.app/expo/issue/ENG-10700/investigate-adding-canceled-conditional-check-for-generic-job-steps

packages/steps/src/BuildWorkflow.ts
szdziedzic: I believe that the first error is usually the true one. The rest of the errors will happen in the cleanup steps which should `always()` be run.

packages/steps/src/BuildStep.ts
szdziedzic: Yeah unfortunately we do, so it is displayed nicely on the website. I also tried with just one line, but it wasn't working.

BTW only `Skipped build step "${this.displayName}"` will be visible in the logs

packages/steps/src/BuildWorkflow.ts
sjchmiela: Exactly. If the "cleanup step" fails, wouldn't we accidentally reassign `maybeError` and forget the first `err`?

szdziedzic: Good catch, sorry, I will fix it

PR #293, opened 2023-11-21
[eas-build-job] Remove `buildMode` from metadata

sjchmiela:
# Why

At the moment we have two `buildMode`s: in metadata and in job. Let's remove one — the one in metadata.

I won't remove `metadata->>'buildMode'` from `www`. The flow will go like this:
- clients specify `job.mode`
- in `www` we assign `metadata.buildMode = job.mode`
- we replace all `metadata.buildMode` in `turtle-v2` with `job.mode`
- we send `job.mode` to `turtle-api`, but not `metadata.buildMode`

# How

Removed `buildMode` from metadata definition.

# Test Plan

None.


packages/eas-build-job/src/metadata.ts
```diff
@@ -198,7 +193,6 @@ export const MetadataSchema = Joi.object({
   message: Joi.string().max(1024),
   runFromCI: Joi.boolean(),
   runWithNoWaitFlag: Joi.boolean(),
-  buildMode: Joi.string().valid('build', 'resign', 'custom'),
```
szdziedzic, line 201: It will break builds is turtle-api https://github.com/expo/turtle-v2/blob/6fd575033d373ab443138a7552e04a767a0672c9/src/services/turtle-api/src/api/builds.ts#L500

szdziedzic (changes requested):

szdziedzic (approved):

packages/eas-build-job/src/metadata.ts
sjchmiela: For future readers — it won't because `stripUnknown: true` https://github.com/expo/eas-build/blob/23e76f86b6152596ad28eb798a64ff3f50fdc212/packages/eas-build-job/src/metadata.ts#L205-L209

PR #294, opened 2023-11-21
[steps] make `null` build step invalid in Joi schema

szdziedzic:
# Why

If you had config like
```yml
build:
  name: Demo
  steps:
    - eas/checkout
    -
```

In CLI you would get 
Screenshot 2023-11-21 at 16 44 14

# How

Make `null` invalid.

# Test Plan

Screenshot 2023-11-21 at 17 03 44

### DIFF
--- a/packages/steps/src/BuildConfig.ts
+++ b/packages/steps/src/BuildConfig.ts
@@ -391,15 +391,15 @@ export function mergeConfigWithImportedFunctions(
 }
 
 export function isBuildStepCommandRun(step: BuildStepConfig): step is BuildStepCommandRun {
-  return step !== null && typeof step === 'object' && typeof step.run === 'object';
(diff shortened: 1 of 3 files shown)


szdziedzic:
BTW `typeof null` is `object`

packages/steps/src/BuildConfig.ts
```diff
@@ -390,15 +391,15 @@ export function mergeConfigWithImportedFunctions(
 }
 
 export function isBuildStepCommandRun(step: BuildStepConfig): step is BuildStepCommandRun {
-  return typeof step === 'object' && typeof step.run === 'object';
+  return step !== null && typeof step === 'object' && typeof step.run === 'object';
```
sjchmiela, line 394: Haven't checked — can `step` be undefined? Should we loosen the comparison to `!=` or `Boolean(step)`?

sjchmiela (approved):

PR #295, opened 2023-11-21
[steps] make generic job workflow always throw first error

szdziedzic:
# Why

https://github.com/expo/eas-build/pull/292#discussion_r1401027194

# How

Always throw the first error when running generic job workflow

# Test Plan

Add test

### DIFF
--- a/packages/steps/src/BuildWorkflow.ts
+++ b/packages/steps/src/BuildWorkflow.ts
@@ -23,7 +23,9 @@ export class BuildWorkflow {
         try {
           await step.executeAsync();
         } catch (err: any) {
-          maybeError = err;
+          if (!maybeError) {
+            maybeError = err;
+          }
           hasAnyPreviousStepFailed = true;
         }
       } else {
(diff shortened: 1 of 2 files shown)


packages/steps/src/__tests__/BuildWorkflow-test.ts
```diff
@@ -130,5 +130,31 @@ describe(BuildWorkflow, () => {
       verify(mockBuildStep2.executeAsync()).once();
       verify(mockBuildStep3.executeAsync()).once();
     });
+
+    it('throws always the fist error', async () => {
```
szdziedzic, line 134:
```suggestion
    it('throws always the first error', async () => {
```

packages/steps/src/BuildWorkflow.ts
```diff
@@ -23,7 +23,9 @@ export class BuildWorkflow {
         try {
           await step.executeAsync();
         } catch (err: any) {
-          maybeError = err;
+          if (!maybeError) {
+            maybeError = err;
+          }
```
sjchmiela, line 28: I find it cleaner, but I'll let you decide

```suggestion
          maybeError = maybeError ?? err;
```

sjchmiela (approved):

PR #297, opened 2023-11-23
[build-tools] [ENG-10483] Add timeouts for install dependencies phase

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-10483/investigate-why-a-pair-of-builds-froze-during-yarn-install
https://linear.app/expo/issue/ENG-10483/investigate-why-a-pair-of-builds-froze-during-yarn-install#comment-7fa0acad

# How

Adds two timeouts to the install dependencies phase. First executes after 15 minutes and leaves a warning message in the logs for the user to check their package.json. Second executes after 30 minutes leaving an error message and throwing an error. If phase ends before the timeouts they are cleared

# Test Plan

Tested by running www and turtle (with the local version of build-tools) locally and running iOS and Android builds

### DIFF
--- a/packages/build-tools/src/common/setup.ts
+++ b/packages/build-tools/src/common/setup.ts
@@ -21,6 +21,8 @@ import { installDependenciesAsync, resolvePackagerDir } from './installDependenc
 import { configureEnvFromBuildProfileAsync, runEasBuildInternalAsync } from './easBuildInternal';
 
 const MAX_EXPO_DOCTOR_TIMEOUT_MS = 30 * 1000;
+const INSTALL_DEPENDENCIES_WARN_TIMEOUT_MS = 15 * 60 * 1000;
+const INSTALL_DEPENDENCIES_KILL_TIMEOUT_MS = 30 * 60 * 1000;
 
 class DoctorTimeoutError extends Error {}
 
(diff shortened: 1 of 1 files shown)


szdziedzic (approved):
seems right

radoslawkrzemien:
@szdziedzic Please have another look after adding the changes we discussed on Friday 🙏

brentvatne (commented):
i thought about this a bit more after our call today and the following occurred to me:

- the problem here is not just that the build hangs, but also that it hangs with no output to logs.
- ideally, we'd start our timeout counter after logs are updated rather than the start of the phase. so, using the current thresholds, a timeout would be considered 30 minutes after the most recent log is printed, and a warning would be 15 minutes after the most recent log is printed. i think this would actually help us lower our thresholds too, reducing the probability of us accidentally canceling a build that just has an unusually long dependency installation phase. 

wdyt?

radoslawkrzemien:
> i thought about this a bit more after our call today and the following occurred to me:
> 
> * the problem here is not just that the build hangs, but also that it hangs with no output to logs.
> * ideally, we'd start our timeout counter after logs are updated rather than the start of the phase. so, using the current thresholds, a timeout would be considered 30 minutes after the most recent log is printed, and a warning would be 15 minutes after the most recent log is printed. i think this would actually help us lower our thresholds too, reducing the probability of us accidentally canceling a build that just has an unusually long dependency installation phase.
> 
> wdyt?

Sounds like a really good solution, I'm just not sure how to implement it. But I'll look into it some more

radoslawkrzemien:
@brentvatne 
Got it work like that (disregard the 1.5/3 seconds, this was to shorten the timeouts to trigger the conditions)
Screenshot 2023-11-28 at 14 24 30

The timeouts get reset after every line is printed (as you can see the warning happened twice because of that), so they should trigger not after a set interval, but after a *set interval of no activity (no logs)*

szdziedzic (approved):
Looks reasonable

PR #298, opened 2023-11-28
[logger] [ENG-10483] Add callback

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-10483/investigate-why-a-pair-of-builds-froze-during-yarn-install
https://linear.app/expo/issue/ENG-10483/investigate-why-a-pair-of-builds-froze-during-yarn-install#comment-7fa0acad
https://github.com/expo/eas-build/pull/297#pullrequestreview-1751696608

Companion of:
- https://github.com/expo/eas-build/pull/297
- https://github.com/expo/eas-build/pull/299

# How

Added an optional callback to the function piping spawn output into the child logger. This callback gets executed whenever the child logger calls the info() function to print new log line. In this case it is needed to refresh a timeout so it gets reset after each line from spawn process

# Test Plan

Tested by running builds locally with passed callback to reset timeouts and it worked as expected

# Deployment plan

1. `logger`
2. `turtle-spawn`
3. `build-tools`
4. `turtle`

### DIFF
--- a/packages/logger/src/pipe.ts
+++ b/packages/logger/src/pipe.ts
@@ -38,7 +38,8 @@ export enum PipeMode {
 function pipeSpawnOutput(
   logger: bunyan,
   { stdout, stderr }: SpawnOutput = {},
-  { mode = PipeMode.COMBINED, lineTransformer }: PipeOptions = {}
+  { mode = PipeMode.COMBINED, lineTransformer }: PipeOptions = {},
+  infoCallbackFn?: () => void
 ): void {
   if (stdout && [PipeMode.COMBINED, PipeMode.COMBINED_AS_STDOUT].includes(mode)) {
     const stdoutLogger = logger.child({ source: 'stdout' });
@@ -46,6 +47,9 @@ function pipeSpawnOutput(
       stdout,
       (line) => {
         stdoutLogger.info(line);
+        if (infoCallbackFn) {
+          infoCallbackFn();
+        }
       },
       lineTransformer
     );
@@ -60,6 +64,9 @@ function pipeSpawnOutput(
       stderr,
       (line) => {
         stderrLogger.info(line);
+        if (infoCallbackFn) {
+          infoCallbackFn();
+        }
       },
       lineTransformer
     );
--- a/packages/logger/src/pipe.ts
+++ b/packages/logger/src/pipe.ts
@@ -9,7 +9,7 @@ interface SpawnOutput {
   stderr?: Readable | null;
 }
 
-interface PipeOptions {
+export interface PipeOptions {
   mode?: PipeMode;
   lineTransformer?: LineTransformer;
   infoCallbackFn?: () => void;
--- a/packages/logger/src/pipe.ts
+++ b/packages/logger/src/pipe.ts
@@ -12,6 +12,7 @@ interface SpawnOutput {
 interface PipeOptions {
   mode?: PipeMode;
   lineTransformer?: LineTransformer;
+  infoCallbackFn?: () => void;
 }
 
 function pipe(stream: Readable, loggerFn: LineLogger, lineTransformer?: LineTransformer): void {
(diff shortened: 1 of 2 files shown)


packages/logger/src/pipe.ts
```diff
@@ -60,6 +64,9 @@ function pipeSpawnOutput(
       stderr,
       (line) => {
         stderrLogger.info(line);
+        if (infoCallbackFn) {
+          infoCallbackFn();
+        }
```
sjchmiela, line 69:
```suggestion
        infoCallbackFn?.();
```
how about ^

sjchmiela (commented):
Could we have reused `lineTransformer`?

packages/logger/src/pipe.ts
```diff
@@ -38,14 +38,18 @@ export enum PipeMode {
 function pipeSpawnOutput(
   logger: bunyan,
   { stdout, stderr }: SpawnOutput = {},
-  { mode = PipeMode.COMBINED, lineTransformer }: PipeOptions = {}
+  { mode = PipeMode.COMBINED, lineTransformer }: PipeOptions = {},
```
sjchmiela, line 41: If adding, why not add it to `PipeOptions`? Maybe then https://github.com/expo/eas-build/pull/299 wouldn't even be necessary?

packages/logger/src/pipe.ts
radoslawkrzemien: Looks good

packages/logger/src/pipe.ts
radoslawkrzemien: Would still need to update `SpawnOptions` in `turtle-spawn/src/index.ts` if I'm not mistaken

radoslawkrzemien:
> Could we have reused `lineTransformer`?

Hm, not sure, but I will check

packages/logger/src/pipe.ts
sjchmiela: 👍 Wouldn't it be cooler to export `PipeOptions` and `SpawnOutput` from `logger` and import (& re-export) in `turtle-spawn`?

radoslawkrzemien: Hm, I think I'm tired, but I don't see what problem would that solve. Can you elaborate on that?

sjchmiela: `SpawnOptions` defined manually in `turtle-spawn` https://github.com/expo/eas-build/blob/01defae0efe516a2378bf8375a11842457e8b05f/packages/turtle-spawn/src/index.ts#L8-L12 merge `SpawnAsyncOptions` (that is ok to me) and manually: `logger`, `lineTransformer` and `mode`. The `logger` is indeed used by the `spawn`, but `lineTransformer` and `mode` are closely (and only) related to `pipeSpawnOutput`.

Unless I don't see something, there's no reason to define `lineTransformer` and `mode` manually in `turtle-spawn`. Instead we should export `PipeOptions` and define `SpawnOptions = SpawnAsyncOptions & PipeOptions & { logger }`. This way no types need to be changed if we add/remove options from either of the packages?

radoslawkrzemien: Ohhh, right. Yeah that does sound reasonable

radoslawkrzemien:
> > Could we have reused `lineTransformer`?
> 
> Hm, not sure, but I will check

I think we could, but I'm not sure if I want to. `lineTransformer` sounds like it is meant to modify what is printed, not to leave it as-is and do something else. Because of that, I would rather add a callback function that clearly has nothing to do with what is printed, just executes something whenever printing happens. Could be more confusing for someone else otherwise

radoslawkrzemien:
> > > Could we have reused `lineTransformer`?
> > 
> > 
> > Hm, not sure, but I will check
> 
> I think we could, but I'm not sure if I want to. `lineTransformer` sounds like it is meant to modify what is printed, not to leave it as-is and do something else. Because of that, I would rather add a callback function that clearly has nothing to do with what is printed, just executes something whenever printing happens. Could be more confusing for someone else otherwise

Besides, if `lineTransformer` returns an empty line it does not get printed, but the transformer executes (`pipe.ts:72`). Because of that there could be a situation where nothing gets printed but the timeouts get refreshed anyway

szdziedzic (approved):

PR #300, opened 2023-12-05
[ENG-10477][steps][build-tools] actually use correct default working directory if project wasn't checked out

szdziedzic:
# Why

https://github.com/expo/eas-build/pull/291 was reverted because https://github.com/expo/eas-build/pull/296

Actually fix the issue and revert the revert 

# How

The working dir for the step was determined when creating the step (at the beginning of processing custom workflow) so changing the `defaultWorkingDir` during the process wouldn't have any effect.

I made workingDir to be resolved in getter based on defaultWorkingDir every time

# Test Plan

Tests

Run the build which failed on staging locally and see that the path is correct

### DIFF
--- a/packages/build-tools/src/steps/functions/checkout.ts
+++ b/packages/build-tools/src/steps/functions/checkout.ts
@@ -15,6 +15,7 @@ export function createCheckoutBuildFunction(): BuildFunction {
           overwrite: true,
         }
       );
+      stepsCtx.global.markAsCheckedOut();
     },
   });
 }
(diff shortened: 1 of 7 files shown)


khamilowicz:
Looks good,  but changing location of `workingDirectory` during workflow execution and conditional on specific step doesn't sit well with me. I'm afraid it may cause confusion.
Maybe, in `markAsCheckedOut`, we can add a log line with something like `project checked out, changing working directory to x`, just to make it obvious what and why happened.

khamilowicz (approved):

radoslawkrzemien (approved):

PR #302, opened 2024-01-04
Remove support for S3 archive type

sjchmiela:
# Why

Since https://github.com/expo/universe/pull/14139 we don't support uploading projects to S3.

# How

Removed `ArchiveSourceType.S3` and its dependents.

# Test Plan

`yarn build` in root passed.


szdziedzic (commented):
Maybe let's remove https://github.com/expo/eas-cli/blob/2dab7846bebef4d078b401855ecd3a1747b2c2a9/packages/eas-cli/src/build/graphql.ts#L24-L29 first

szdziedzic (approved):

PR #306, opened 2024-01-12
Refactor uploading artifacts

sjchmiela:
# Why

This should make adding generic artifacts easier.

# How

- renamed `ArtifactType` to `ManagedArtifactType` and moved to `eas-build-job`. `ManagedArtifactType`s are artifacts recognized and produced by managed workflow — application archive, "build artifacts" and Xcode logs. `GenericArtifactType`s will have much more possibilities.
- changed `uploadArtifacts` to accept a single object argument instead of a list. _Much_ better.

# Test Plan

I tested this with [the generic artifact types changes](https://github.com/expo/eas-build/pull/307). Mostly, those are types changes, so if types are all right, the pull request should be all right.

### DIFF
--- a/packages/eas-build-job/src/artifacts.ts
+++ b/packages/eas-build-job/src/artifacts.ts
@@ -1,0 +1,8 @@
+export enum ManagedArtifactType {
+  APPLICATION_ARCHIVE = 'APPLICATION_ARCHIVE',
+  BUILD_ARTIFACTS = 'BUILD_ARTIFACTS',
+  /**
+   * @deprecated
+   */
+  XCODE_BUILD_LOGS = 'XCODE_BUILD_LOGS',
+}
--- a/packages/local-build-plugin/src/android.ts
+++ b/packages/local-build-plugin/src/android.ts
@@ -28,7 +28,7 @@ export async function buildAndroidAsync(
       if (artifact.type !== ManagedArtifactType.APPLICATION_ARCHIVE) {
         return null;
       } else {
-        return await prepareArtifacts(artifact.paths, logger);
+        return await prepareArtifacts(artifact.paths, artifact.logger);
       }
     },
     env,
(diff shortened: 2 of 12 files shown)


packages/eas-build-job/src/artifacts.ts
```diff
@@ -0,0 +1,8 @@
+export enum ManagedArtifactType {
+  APPLICATION_ARCHIVE = 'APPLICATION_ARCHIVE',
+  BUILD_ARTIFACTS = 'BUILD_ARTIFACTS',
+  /**
+   * @deprecated
+   */
```
szdziedzic, line 6: Do you know why it is deprecated? I know it was `deperecated` before your PR but I'm wondering what the intention was here

szdziedzic (approved):
Looks good, thanks!

packages/eas-build-job/src/artifacts.ts
sjchmiela: I think at some point we stopped uploading Xcode logs as "build artifacts" and started to treat them separately — not only do they land in a separate bucket (`turtle-v2-logs` vs `turtle-v2-artifacts`), but also they don't get renamed as other artifacts do (from `prefix` to `prefix/name.suffix`).

packages/local-build-plugin/src/android.ts
```diff
@@ -25,11 +24,11 @@ export async function buildAndroidAsync(
...
     logBuffer,
     runGlobalExpoCliCommand: runGlobalExpoCliCommandAsync,
-    uploadArtifacts: async (type: ArtifactType, paths: string[], logger?: bunyan) => {
-      if (type !== ArtifactType.APPLICATION_ARCHIVE) {
+    uploadArtifacts: async (artifact: ArtifactToUpload) => {
+      if (artifact.type !== ManagedArtifactType.APPLICATION_ARCHIVE) {
         return null;
       } else {
-        return await prepareArtifacts(paths, logger);
+        return await prepareArtifacts(artifact.paths, logger);
```
sjchmiela, line 31:
```suggestion
        return await prepareArtifacts(artifact.paths, artifact.logger);
```

PR #307, opened 2024-01-12
Add support for generic artifacts

sjchmiela:
# Why

We want to allow users to upload multiple artifacts, possibly of similar meaning (e.g. multiple iOS apps).

# How

Added `GenericArtifactType` alongside `ManagedArtifactType`.

# Test Plan

- [x] Tested generic artifacts with complementary `turtle-v2` changes.

### DIFF
--- a/packages/build-tools/src/context.ts
+++ b/packages/build-tools/src/context.ts
@@ -12,6 +12,8 @@ import {
   errors,
   Metadata,
   EnvironmentSecretType,
+  GenericArtifactType,
+  isGenericArtifact,
 } from '@expo/eas-build-job';
 import { ExpoConfig } from '@expo/config';
 import { bunyan } from '@expo/logger';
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/context.ts
```diff
@@ -219,7 +227,7 @@ export class BuildContext<TJob extends Job> {
     logger: bunyan;
   }): Promise<void> {
     const bucketKey = await this._uploadArtifact({ artifact, logger });
-    if (bucketKey) {
+    if (bucketKey && !isGenericArtifact(artifact)) {
       this.artifacts[artifact.type] = bucketKey;
     }
```
szdziedzic, line 232: Sorry probably this is a stupid question, but where will be the information about the generic artifacts currently stored?

szdziedzic: Oh okay, I think I know the answer, we don't support generic artifacts for old build flow so we don't care about storing info about them here.

szdziedzic (approved):

packages/build-tools/src/context.ts
sjchmiela: My initial ideas are somewhere around:
- we know what type of the application archive is being built for managed builds based on flags (`platform`, `simulator`, maybe others)
- we could upload managed application archives as:
  - generic artifact of respective type (`${platform}_${...}`)
  - _and_ set as `APPLICATION_ARCHIVE` bucket key too
- managed build artifacts are just `UNKNOWN` generic artifacts
- then we move `www` and `website` to use generic artifacts
- then we can remove managed artifacts altogether.

PR #308, opened 2024-01-16
[ENG-11093][build-tools] improve error message for "Run Fastlane" build phase when doing resign

szdziedzic:
# Why

https://exponent-internal.slack.com/archives/C02123T524U/p1705340342478629

# How

Improve error message for `Run Fastlane` build phase when doing resign.

# Test Plan

Tests


packages/build-tools/src/buildErrors/userErrorHandlers.ts
```diff
@@ -253,6 +253,17 @@ To resolve this issue, downgrade to an older Xcode version using the "image" fie
...
   },
+  {
+    platform: Platform.IOS,
+    phase: BuildPhase.RUN_FASTLANE,
+    mode: BuildMode.RESIGN,
+    regexp: /.*/,
+    createError: () =>
+      new UserFacingError(
+        errors.ErrorCode.UNKNOWN_FASTLANE_RESIGN_ERROR,
+        `The "Run fastlane" step failed with an unknown error. Inspect the logs above to verify if the bundle identifier associated with your new provisioning profile matches the bundle identifier of the build you want to resign.`
```
sjchmiela, line 264: Is this the only/main reason resign may fail? Maybe we shouldn't point to bundle identifiers if this may not be the issue?

sjchmiela (approved):

packages/build-tools/src/buildErrors/userErrorHandlers.ts
szdziedzic: TBH I don't have enough data about whether this is the main reason, just assumed it

szdziedzic: Do you suggest only telling people to inspect the logs above?

sjchmiela: Yea.

If we wanted to invest more time into it, we could fetch a couple failed resign jobs from the DB and see what were the reasons. If all turn out to be profile-build mismatch, we could leave it as it is.

radoslawkrzemien: I concur, if there are more common cases why this could fail, such a suggestion could point the user to look in the wrong place and waste time, which some users will surely complain about

radoslawkrzemien (approved):
Looks good 🚀

packages/build-tools/src/buildErrors/userErrorHandlers.ts
szdziedzic: WDYT about it now?

sjchmiela: Cool!

PR #309, opened 2024-01-16
Change `uploadArtifacts` into `uploadArtifact`

sjchmiela:
# Why

We won't "upload artifacts" anymore the way we did. Also, artifacts have types which should live nested inside `artifact` property and that property should live _alongside_ logger.

# How

Renamed functions, changed calls.

# Test Plan

Believing in TS.


szdziedzic (approved):
Looks reasonable, although please fix these warnings 😄 

I believe you need to do `yarn install` again after I bumped all deps

PR #323, opened 2024-01-26
Create `eas/start_ios_simulator` build function

sjchmiela:
# Why

Part of the Maestro example.

# How

Moved my tested EAS workflow into a TS function.

It prints all available simulator in the beginning so the users have a list to compare their request to, if they need to. We'll need to make sure to educate people that Simulator Devices, UDIDs are image-specific.

A better version of this function could accept an iOS version to test on, would download and install necessary runtime and create a dedicated device for tests, but I suspect this is enough for now.

# Test Plan

I tested it a bunch, with Simulator stopped, started, with different inputs (valid, invalid).

One thing I don't like is that stdout and stderr of a `spawn` can lose ordering, see:
Zrzut ekranu 2024-01-26 o 22 14 51
Zrzut ekranu 2024-01-26 o 22 14 38

Is there anything we can do about it? Maybe I'm doing something wrong?

### DIFF
--- a/packages/build-tools/src/steps/functions/startIosSimulator.ts
+++ b/packages/build-tools/src/steps/functions/startIosSimulator.ts
@@ -1,0 +1,149 @@
+import { PipeMode } from '@expo/logger';
+import { BuildFunction, BuildStepInput, BuildStepInputValueTypeName } from '@expo/steps';
+import spawn from '@expo/turtle-spawn';
+import { minBy } from 'lodash';
+
+import { retryAsync } from '../../utils/retry';
+
+export function createStartIosSimulatorBuildFunction(): BuildFunction {
+  return new BuildFunction({
+    namespace: 'eas',
+    id: 'start_ios_simulator',
+    name: 'Start iOS Simulator',
+    inputProviders: [
+      BuildStepInput.createProvider({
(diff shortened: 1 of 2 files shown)


packages/build-tools/src/steps/functions/startIosSimulator.ts
```diff
@@ -0,0 +1,148 @@
...
+async function sleep(timeMs: number): Promise<void> {
+  await new Promise<void>((resolve) => setTimeout(resolve, timeMs));
+}
+
+async function findMostGenericIphone(): Promise<AvailableXcrunSimctlDevice | null> {
+  const availableSimulatorDevices = await getAvailableSimulatorDevices();
+  const availableIphones = availableSimulatorDevices.filter((device) =>
+    device.name.startsWith('iPhone')
+  );
+  // It's funny, but it works.
```
szdziedzic, line 75: interesting 😄

szdziedzic (approved):
Looks reasonable to me

If you want to look for inspiration when it comes to launching simulators/emulators the Expo CLI is a good place to check. 
https://github.com/expo/expo/tree/main/packages/%40expo/cli

I reused some logic from there when working on the `eas build:run` command

PR #324, opened 2024-01-29
Add `eas/start_android_emulator` build function

sjchmiela:
# Why

We want to provide users with an easy way to start an Android emulator for E2E tests.

# How

It's rather simple, just a couple bash commands. I'm happy by using `detached` I was able to work around the ([previously necessary](https://exponent-internal.slack.com/archives/C05JR16NWAX/p1691606736451939)) `echo 'nohup ...' | at now`.

# Test Plan

I've played with it a bunch and even was able to start two emulators at once! The emulators are alive as long as `eas-build-worker` is alive, which I think is ok.

### DIFF
--- a/packages/build-tools/src/steps/functions/startAndroidEmulator.ts
+++ b/packages/build-tools/src/steps/functions/startAndroidEmulator.ts
@@ -16,6 +16,12 @@ export function createStartAndroidEmulatorBuildFunction(): BuildFunction {
     name: 'Start Android Emulator',
     inputProviders: [
       BuildStepInput.createProvider({
+        id: 'device_name',
+        required: false,
+        defaultValue: 'EasAndroidDevice01',
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
+      BuildStepInput.createProvider({
         id: 'system_image_package',
         required: false,
         defaultValue: defaultSystemImagePackage,
@@ -23,7 +29,7 @@ export function createStartAndroidEmulatorBuildFunction(): BuildFunction {
       }),
     ],
     fn: async ({ logger }, { inputs }) => {
-      const deviceName = 'EasAndroidDevice01';
+      const deviceName = `${inputs.device_name.value}`;
       const systemImagePackage = `${inputs.system_image_package.value}`;
       logger.info('Making sure system image is installed');
       await spawn('sdkmanager', [systemImagePackage], {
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/steps/functions/startAndroidEmulator.ts
```diff
@@ -0,0 +1,147 @@
...
+
+      logger.info('Creating emulator device');
+      const avdManager = spawn(
+        'avdmanager',
+        ['create', 'avd', '--name', deviceName, '--package', systemImagePackage, '--force'],
+        {
+          stdio: 'pipe',
+        }
+      );
+      avdManager.child.stdin?.write('no');
```
szdziedzic, line 49: What does it do?

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
```diff
@@ -0,0 +1,147 @@
...
+      let hasBootCompleted = false;
+      while (!hasBootCompleted) {
+        const { stdout } = await spawn(
+          'adb',
+          ['-s', serialId, 'shell', 'getprop', 'sys.boot_completed'],
+          {
+            mode: PipeMode.COMBINED,
+          }
+        );
+        if (stdout.startsWith('1')) {
```
szdziedzic, line 84: Will it always be true?

szdziedzic (approved):
Looks reasonable

As I said previously Expo CLI is a good place to check if you look for inspiration for simulator/emulator utils 😄

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
sjchmiela: It answers `avdmanager`'s question whether we want to create a custom build profile for the AVD. I'll add a comment.

Zrzut ekranu 2024-01-30 o 12 23 37

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
sjchmiela: If the simulator boots successfully, it should, and since we got `qemu.uuid` a second before, the boot should be complete soon too. Are you suggesting changing an infinite loop into a finite `retryAsync`, just in case?

szdziedzic: Yeah I think it would be good to limit the max time it can hang in this loop if you don't mind

sjchmiela: Set to 3 minutes as you did in `eas-cli`.

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
sjchmiela: Added comment.

PR #325, opened 2024-01-30
Add `eas/install_maestro` function

sjchmiela:
# Why

Unfortunately, we don't have all the prerequisites for running Maestro in all our environments. I think it makes sense to let people call `eas/install_maestro` to get Maestro instead of asking them to manually set up Java, IDB when on Mac.

# How

Used previously running workflow where I did those two things. Wrapped in a build function.

# Test Plan

I think _true_ test will only be running in staging or production.


packages/build-tools/src/steps/functions/installMaestro.ts
```diff
@@ -0,0 +1,190 @@
...
+          currentMaestroVersion = await getMaestroVersion();
+        } catch {}
+      }
+
+      if (!currentMaestroVersion || requestedMaestroVersion !== currentMaestroVersion) {
+        await installMaestro({
+          version: requestedMaestroVersion,
+          global,
+          logger,
+        });
```
szdziedzic, line 63: I'm hesitant whether we want to run it for people running custom builds locally (`eas build --local`) 🤔 

I mean I see a case where it can alter software installed on the user's machine, without the user wanting this software to be altered (even though they called the `install_maestro` function). I believe we should do the same thing as for Java here 🤔  

So ultimately this step should not install anything when running in local mode, but rather only tell people what they need to install manually and validate installed stuff in my opinion. 

If this would be something interactive it would be probably best to just prompt the user if they want to install XYZ, but we can't do stuff like this here. So I think it's better to not install software on the user's computer at all vs install it without the user truly wanting to have any new software installed on their device by us.

Maybe to make it more user-friendly we should give people a command they can run to install stuff in an error message.

sjchmiela: That's a reasonable concern. I'll make it so.

sjchmiela: https://github.com/expo/eas-build/blob/a2b2b25659139ce8b6f1c477d02182d2a8a55f78/packages/build-tools/src/steps/functions/installMaestro.ts#L31-L69

packages/build-tools/src/steps/functions/installMaestro.ts
```diff
@@ -0,0 +1,228 @@
...
+      if (env.EAS_BUILD_RUNNER !== 'eas-build') {
+        const currentIsJavaInstalled = await isJavaInstalled();
+        const currentIsIdbInstalled = await isIdbInstalled();
+
+        if (!currentIsJavaInstalled) {
+          logger.warn(
+            'It seems Java is not installed. It is required to run Maestro. If the job fails, this may be the reason.'
+          );
+          logger.info('');
+        }
```
szdziedzic, line 41: Thanks!

szdziedzic (approved):

PR #330, opened 2024-02-06
[eas-build] [ENG-11196] Custom build send Slack message function

radoslawkrzemien:
# Why

# How

Adds new function `eas/send_slack_message` allowing for sending Slack messages to an established Slack webhook URL
For Slack webhook URL customisation added `env` to the static context used for input value interpolation

# Test Plan

Added automated tests

### Slack webhook config
Screenshot 2024-02-06 at 16 56 36

#### With default url set in SLACK_HOOK_URL secret
Screenshot 2024-02-07 at 14 43 38

#### With specific url provided as plain input
Screenshot 2024-02-08 at 16 44 20
#### With specific url provided as a reference to another secret
Screenshot 2024-02-08 at 16 44 36

### Build logs
Screenshot 2024-02-07 at 11 26 56

### Received message
Screenshot 2024-02-07 at 11 23 40

### When Slack webhook URL not provided (either SLACK_HOOK_URL secret not set, or slack_hook_url input not provided)
Screenshot 2024-02-08 at 16 49 08


packages/build-tools/src/steps/functions/sendSlackMessage.ts
```diff
@@ -0,0 +1,64 @@
...
+  return new BuildFunction({
+    namespace: 'eas',
+    id: 'send_slack_message',
+    name: 'Send Slack message',
+    inputProviders: [
+      BuildStepInput.createProvider({
+        id: 'slack_hook_url',
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+        required: true,
+      }),
```
szdziedzic, line 19: I think that currently AFAIK you can't reference secrets in function inputs 🤔 

What if somebody considers it sensitive data and doesn't want to put it as plain text in the config file? 

I'm suggesting here that maybe we should also read this value from `SLACK_WEBHOOK_URL` env var/secret?

packages/build-tools/src/steps/functions/sendSlackMessage.ts
```diff
@@ -0,0 +1,64 @@
...
+  try {
+    fetchResult = await fetch(slackHookUrl, {
+      method: 'POST',
+      body: JSON.stringify(body),
+      headers: { 'Content-Type': 'application/json' },
+    });
+  } catch (error) {
+    logger.warn(`Sending Slack message to webhook url "${slackHookUrl}" failed`);
+    logger.debug(error);
+    throw new Error(`Sending Slack message to webhook url "${slackHookUrl}" failed`);
```
szdziedzic, line 52: Can users see this debug log? If not maybe we should log the underlying error better 🤔? 

Can this error contain secret information we don't want to put in the logs? I don't think so. If somebody considers the webhook URL a secret it will be masked anyway.

packages/build-tools/src/steps/functions/sendSlackMessage.ts
radoslawkrzemien: That may be a good idea

packages/build-tools/src/steps/functions/sendSlackMessage.ts
radoslawkrzemien: Haven't tested it yet, I'll check

packages/build-tools/src/steps/functions/sendSlackMessage.ts
radoslawkrzemien: Changed plain input into a secret

packages/build-tools/src/steps/functions/sendSlackMessage.ts
radoslawkrzemien: User sees the `warn` and the `error`, not the `debug`. I thought of it more as a way for us to find the true error in the logs if ever there will be user asking why his slack messaging doesn't work. But if you think exposing the contents of the error in their entirety is safe, I can do that too

packages/build-tools/src/steps/functions/__tests__/sendSlackMessage.test.ts
```diff
@@ -0,0 +1,227 @@
...
+describe(createSendSlackMessageFunction, () => {
+  let fetchMock: jest.SpyInstance;
+  let loggerInfoMock: jest.SpyInstance;
+  let loggerWarnMock: jest.SpyInstance;
+  let loggerDebugMock: jest.SpyInstance;
+  let loggerErrorMock: jest.SpyInstance;
+
+  beforeEach(() => {
+    fetchMock = jest.mocked(fetch);
+  });
```
sjchmiela, line 19: Do we need to do this in `beforeEach`? It seems a single `const fetchMock` should be sufficient: https://github.com/expo/turtle-v2/blob/917135accef42ab4be498eba2439d63544fa4938/src/libs/turtle-common/src/__unit__/turtleFetch.test.ts#L8

packages/build-tools/src/steps/functions/__tests__/sendSlackMessage.test.ts
```diff
@@ -0,0 +1,227 @@
...
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
```
sjchmiela, line 23: Do we need to do this?

packages/build-tools/src/steps/functions/sendSlackMessage.ts
```diff
@@ -0,0 +1,65 @@
...
+        required: true,
+      }),
+    ],
+    fn: async (stepCtx, { inputs, env }) => {
+      await sendSlackMessageAsync(stepCtx, { inputs, env });
+    },
+  });
+}
+
+export async function sendSlackMessageAsync(
```
sjchmiela, line 29:
```suggestion
async function sendSlackMessageAsync(
```

packages/build-tools/src/steps/functions/__tests__/sendSlackMessage.test.ts
```diff
@@ -0,0 +1,227 @@
...
+      { marker: 'start-step' },
+      'Executing build step "Send Slack message"',
+    ]);
+    expect(loggerInfoMock.mock.calls[1]).toEqual(['Sending Slack message']);
+    expect(loggerInfoMock.mock.calls[2]).toEqual(['Slack message sent successfully']);
+    expect(loggerInfoMock.mock.calls[3]).toEqual([
+      { marker: 'end-step', result: 'success' },
+      'Finished build step "Send Slack message" successfully',
+    ]);
+    expect(loggerWarnMock).not.toHaveBeenCalled();
```
sjchmiela, line 60: I don't think we need to test logs to this level of detail. Same with `start-step` and `end-step` — those are not related to business logic of Slack message function so there's no reason to assert it here.

packages/build-tools/src/steps/functions/sendSlackMessage.ts
```diff
@@ -0,0 +1,65 @@
...
+    ],
+    fn: async (stepCtx, { inputs }) => {
+      await sendSlackMessageAsync(stepCtx, inputs);
+    },
+  });
+}
+
+export async function sendSlackMessageAsync(
+  stepCtx: BuildStepContext,
+  inputs: BuildStepInputById
```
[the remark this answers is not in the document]
radoslawkrzemien: Good idea

packages/build-tools/src/steps/functions/__tests__/sendSlackMessage.test.ts
radoslawkrzemien: Yeah, it's a leftover from the time used builtin `fetch` before I realised that node@16 doesn't have it and I need to use `node-fetch` anyway

packages/build-tools/src/steps/functions/__tests__/sendSlackMessage.test.ts
radoslawkrzemien: Leftover again

packages/build-tools/src/steps/functions/sendSlackMessage.ts
```diff
@@ -0,0 +1,73 @@
...
+  slackMessage: string;
+}): Promise<void> {
+  if (!slackHookUrl) {
+    logger.warn(
+      'Slack webhook URL not provided - provide input "slack_hook_url" or set "SLACK_HOOK_URL" secret'
+    );
+    throw new Error(
+      'Sending Slack message failed - provide input "slack_hook_url" or set "SLACK_HOOK_URL" secret'
+    );
+  }
```
sjchmiela, line 47: nit: My vote would be to put the validation in the `BuildFunction`, so `sendSlackMessageAsync` is only concerned about business and it assumes valid input. Not that it matters here, just my opinion.

PR #333, opened 2024-02-07
[steps][build-tools] add `eas-build` build function group

szdziedzic:
# Why

We want to add a concept called functions group to custom builds, so it is easier for users to just run our build process if they don't want to interfere with it but instead do some stuff after/before the build process.

# How

Add `eas/build` function group. This function group spawns build steps based on `eas` namespace functions required to complete the build in all needed configurations.

Usage:
```yml
build:
  name: test eas/build function group
  steps:
    - run: echo "Hello, world!"
    - eas/build
    - run: echo "Bye, world!"
```
Will do:
1. echo "Hello, world!"
2. checkout
3. install `node_modules`
...
n. upload artifacts
n+1. echo "Bye, world!"

# Test Plan

image
image

TBA Automatic tests

### DIFF
--- a/packages/build-tools/src/steps/functionGroups/build.ts
+++ b/packages/build-tools/src/steps/functionGroups/build.ts
@@ -1,0 +1,149 @@
+import { BuildFunctionGroup, BuildStep, BuildStepGlobalContext } from '@expo/steps';
+import { Platform } from '@expo/eas-build-job';
+
+import { createCheckoutBuildFunction } from '../functions/checkout';
+import { createInstallNodeModulesBuildFunction } from '../functions/installNodeModules';
+import { createPrebuildBuildFunction } from '../functions/prebuild';
+import { createInstallPodsBuildFunction } from '../functions/installPods';
+import { configureEASUpdateIfInstalledFunction } from '../functions/configureEASUpdateIfInstalled';
+import { generateGymfileFromTemplateFunction } from '../functions/generateGymfileFromTemplate';
+import { runFastlaneFunction } from '../functions/runFastlane';
+import { createFindAndUploadBuildArtifactsBuildFunction } from '../functions/findAndUploadBuildArtifacts';
+import { CustomBuildContext } from '../../customBuildContext';
+import { createGetCredentialsForBuildTriggeredByGithubIntegration } from '../functions/getCredentialsForBuildTriggeredByGitHubIntegration';
+import { resolveAppleTeamIdFromCredentialsFunction } from '../functions/resolveAppleTeamIdFromCredentials';
+import { configureIosCredentialsFunction } from '../functions/configureIosCredentials';
+import { runGradleFunction } from '../functions/runGradle';
+import { configureIosVersionFunction } from '../functions/configureIosVersion';
+import { injectAndroidCredentialsFunction } from '../functions/injectAndroidCredentials';
+import { configureAndroidVersionFunction } from '../functions/configureAndroidVersion';
+
+interface HelperFunctionsInput {
+  globalCtx: BuildStepGlobalContext;
+  buildToolsContext: CustomBuildContext;
+}
+
+export function createEasBuildBuildFunctionGroup(
+  buildToolsContext: CustomBuildContext
+): BuildFunctionGroup {
+  return new BuildFunctionGroup({
+    namespace: 'eas',
+    id: 'build',
+    createBuildStepsFromFunctionGroupCall: (globalCtx) => {
+      if (buildToolsContext.job.platform === Platform.IOS) {
+        // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+        if (buildToolsContext.job.simulator || !buildToolsContext.job.secrets?.buildCredentials) {
+          return createStepsForIosSimulatorBuild({
+            globalCtx,
+            buildToolsContext,
+          });
+        } else {
+          return createStepsForIosBuildWithCredentials({
+            globalCtx,
+            buildToolsContext,
+          });
+        }
+      } else {
+        if (!buildToolsContext.job.secrets?.buildCredentials) {
+          return createStepsForAndroidBuildWithoutCredentials({
+            globalCtx,
+            buildToolsContext,
+          });
+        } else {
+          return createStepsForAndroidBuildWithCredentials({
+            globalCtx,
+            buildToolsContext,
+          });
+        }
+      }
+    },
+  });
+}
+
+function createStepsForIosSimulatorBuild({
+  globalCtx,
+  buildToolsContext,
+}: HelperFunctionsInput): BuildStep[] {
+  return [
+    createCheckoutBuildFunction().createBuildStepFromFunctionCall(globalCtx),
+    createInstallNodeModulesBuildFunction().createBuildStepFromFunctionCall(globalCtx),
+    createPrebuildBuildFunction().createBuildStepFromFunctionCall(globalCtx),
+    createInstallPodsBuildFunction().createBuildStepFromFunctionCall(globalCtx),
+    configureEASUpdateIfInstalledFunction().createBuildStepFromFunctionCall(globalCtx),
+    generateGymfileFromTemplateFunction().createBuildStepFromFunctionCall(globalCtx),
+    runFastlaneFunction().createBuildStepFromFunctionCall(globalCtx),
+    createFindAndUploadBuildArtifactsBuildFunction(
+      buildToolsContext
+    ).createBuildStepFromFunctionCall(globalCtx),
+  ];
+}
+
+function createStepsForIosBuildWithCredentials({
+  globalCtx,
+  buildToolsContext,
+}: HelperFunctionsInput): BuildStep[] {
+  const resolveAppleTeamIdFromCredentials =
+    resolveAppleTeamIdFromCredentialsFunction().createBuildStepFromFunctionCall(globalCtx, {
+      id: 'resolve_apple_team_id_from_credentials',
+    });
+  const prebuildStep = createPrebuildBuildFunction().createBuildStepFromFunctionCall(globalCtx, {
+    callInputs: {
+      apple_team_id: '${ steps.resolve_apple_team_id_from_credentials.apple_team_id }',
+    },
+  });
+  return [
+    createCheckoutBuildFunction().createBuildStepFromFunctionCall(globalCtx),
+    createInstallNodeModulesBuildFunction().createBuildStepFromFunctionCall(globalCtx),
+    createGetCredentialsForBuildTriggeredByGithubIntegration(
+      buildToolsContext
+    ).createBuildStepFromFunctionCall(globalCtx),
+    resolveAppleTeamIdFromCredentials,
+    prebuildStep,
+    createInstallPodsBuildFunction().createBuildStepFromFunctionCall(globalCtx),
+    configureEASUpdateIfInstalledFunction().createBuildStepFromFunctionCall(globalCtx),
+    configureIosCredentialsFunction().createBuildStepFromFunctionCall(globalCtx),
+    configureIosVersionFunction().createBuildStepFromFunctionCall(globalCtx),
+    generateGymfileFromTemplateFunction().createBuildStepFromFunctionCall(globalCtx),
+    runFastlaneFunction().createBuildStepFromFunctionCall(globalCtx),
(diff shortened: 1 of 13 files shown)


packages/steps/src/BuildConfigParser.ts
```diff
@@ -126,14 +153,43 @@ export class BuildConfigParser {
...
     return buildFunction.createBuildStepFromFunctionCall(this.ctx);
   }
 
+  private createBuildStepsFromBuildStepFunctionOrBuildStepFunctionGroupCall(
+    buildFunctions: BuildFunctionById,
+    buildFunctionGroups: BuildFunctionGroupById,
+    functionOrFunctionGroupId: string
+  ): BuildStep[] {
+    const maybeFunctionGroup = buildFunctionGroups[functionOrFunctionGroupId];
+    if (maybeFunctionGroup !== undefined) {
```
sjchmiela, line 182:
```suggestion
    if (maybeFunctionGroup) {
```

packages/steps/src/BuildConfigParser.ts
```diff
@@ -126,14 +153,43 @@ export class BuildConfigParser {
...
   private createBuildStepFromBuildStepBareFunctionCall(
     buildFunctions: BuildFunctionById,
-    functionId: BuildStepBareFunctionCall
+    functionId: BuildStepBareFunctionOrFunctionGroupCall
   ): BuildStep {
     const buildFunction = buildFunctions[functionId];
     return buildFunction.createBuildStepFromFunctionCall(this.ctx);
   }
 
+  private createBuildStepsFromBuildStepFunctionOrBuildStepFunctionGroupCall(
```
sjchmiela, line 176:
```suggestion
  private createBuildStepsFromBareBuildStepFunctionOrBareBuildStepFunctionGroupCall(
```
?

packages/steps/src/BuildConfigParser.ts
```diff
@@ -31,50 +31,77 @@ import { BuildWorkflowValidator } from './BuildWorkflowValidator.js';
...
+      return [this.createBuildStepFromBuildStepBareCommandRun(buildStepConfig)];
+    } else if (isBuildStepBareFunctionOrFunctionGroupCall(buildStepConfig)) {
+      return this.createBuildStepsFromBuildStepFunctionOrBuildStepFunctionGroupCall(
+        buildFunctions,
+        buildFunctionGroups,
+        buildStepConfig
+      );
     } else if (buildStepConfig !== null) {
-      return this.createBuildStepFromBuildStepFunctionCall(buildFunctions, buildStepConfig);
+      return [this.createBuildStepFromBuildStepFunctionCall(buildFunctions, buildStepConfig)];
```
sjchmiela, line 104: Should we allow inputs to function groups?

packages/build-tools/src/steps/functions/configureIosVersion.ts
```diff
@@ -64,6 +64,13 @@ export function configureIosVersionFunction(): BuildFunction {
         );
       }
 
+      if (!buildNumber && !appVersion) {
```
sjchmiela, line 67: Should we allow passing inputs to function groups and make the inputs available to all functions inside? Icky or necessary? (Honest question) How will `eas/build` work in a monorepo?

packages/build-tools/src/steps/functionGroups/build.ts
```diff
@@ -0,0 +1,149 @@
...
+import { configureIosVersionFunction } from '../functions/configureIosVersion';
+import { injectAndroidCredentialsFunction } from '../functions/injectAndroidCredentials';
+import { configureAndroidVersionFunction } from '../functions/configureAndroidVersion';
+
+interface HelperFunctionsInput {
+  globalCtx: BuildStepGlobalContext;
+  buildToolsContext: CustomBuildContext;
+}
+
+export function createEasBuildBuildFunctionGroup(
```
sjchmiela, line 26: And _now_ we can delete the non-custom builds flow haha!

packages/build-tools/src/steps/functions/installPods.ts
```diff
@@ -0,0 +1,18 @@
...
+  return new BuildFunction({
+    namespace: 'eas',
+    id: 'install_pods',
+    name: 'Install Pods',
+    fn: async (stepsCtx, { env }) => {
+      stepsCtx.logger.info('Installing pods');
+      await spawn('pod', ['install'], {
+        stdio: 'pipe',
+        env,
+        cwd: stepsCtx.workingDirectory,
```
sjchmiela, line 14: Shouldn't this execute in `ios` or something? What about monorepos?

packages/build-tools/src/steps/functions/installPods.ts
```diff
@@ -0,0 +1,18 @@
...
+import spawn from '@expo/turtle-spawn';
+
+export function createInstallPodsBuildFunction(): BuildFunction {
+  return new BuildFunction({
+    namespace: 'eas',
+    id: 'install_pods',
+    name: 'Install Pods',
+    fn: async (stepsCtx, { env }) => {
+      stepsCtx.logger.info('Installing pods');
+      await spawn('pod', ['install'], {
```
sjchmiela, line 11: Should we use the `npx pod-install` thing? https://linear.app/expo/issue/ENG-10529/discuss-replace-pod-install-step-with-npx-pod-installlatest-on-eas

packages/build-tools/src/steps/utils/expoUpdates.ts
```diff
@@ -35,9 +35,8 @@ export async function configureEASUpdateIfInstalledAsync({
   const expoUpdatesPackageVersion =
     await getExpoUpdatesPackageVersionIfInstalledAsync(workingDirectory);
   if (expoUpdatesPackageVersion === null) {
-    throw new Error(
-      `Cannot configure Expo Updates because the expo-updates package is not installed.`
-    );
+    logger.info('Expo Updates is not installed, skipping configuring Expo Updates.');
```
sjchmiela, line 38: Should we have some internal parameter that allows this to skip this only if run from within a function group? Like, if a user tries to configure updates in a manually written workflow, should we error?

sjchmiela: If the function group could dynamically define next step it would be ideal, but much more complicated.

packages/steps/src/BuildConfigParser.ts
szdziedzic: Maybe in the future, in this PR I would keep it as concise as possible

packages/build-tools/src/steps/functions/configureIosVersion.ts
szdziedzic: TBH I'm not sure if this is necessary for v1, `eas/build` by default operates on "inputs" from `job` (if other inputs are not provided).

szdziedzic: >How will eas/build work in a monorepo
I hope it will work fine xD, I believe that our `eas/install_node_modules` function works well with monorepos

packages/build-tools/src/steps/functionGroups/build.ts
szdziedzic: hahah xD

packages/build-tools/src/steps/functions/installPods.ts
szdziedzic: It should

sjchmiela: What I meant by mentioning monorepos is that if I have a monorepo where the Expo app is under `packages/mobile-app` or something, things will go bad.

I think it's ok though for v1, as you suggest. Let's let users build normal apps with `eas/build` and let the more advanced ones use specific steps from examples.

szdziedzic: If we do working directory = `./ios` it will be run in the correct directory because after the `checkout` the `.` resolves to react native project directory

szdziedzic: So it should also work for monorepos

szdziedzic: Fixing it

packages/build-tools/src/steps/functions/installPods.ts
szdziedzic: I think we should keep it 1:1 with EAS Build standard flow which is currently using `pod install`

packages/build-tools/src/steps/utils/expoUpdates.ts
szdziedzic: yeah, I agree

szdziedzic: Added new input to solve for this

packages/steps/src/BuildConfigParser.ts
```diff
@@ -126,14 +154,43 @@ export class BuildConfigParser {
     });
   }
 
+  private createBuildStepsFromBuildStepBareFunctionGroupCall(
+    buildFunctionGroups: BuildFunctionGroupById,
+    functionGroupId: string
+  ): BuildStep[] {
+    const buildFunctionGroup = buildFunctionGroups[functionGroupId];
+    assert(
+      buildFunctionGroup !== undefined,
```
sjchmiela, line 163:
```suggestion
      buildFunctionGroup,
```

packages/steps/src/BuildConfig.ts
```diff
@@ -456,3 +470,7 @@ function maybeResolveCustomFunctionRelativePath(dir: string, customFunctionPath:
   }
   return customFunctionPath;
 }
+
+function isFulldIdNamespaced(fullId: string): boolean {
```
sjchmiela, line 474:
```suggestion
function isFullIdNamespaced(fullId: string): boolean {
```

packages/steps/src/BuildConfig.ts
```diff
@@ -410,42 +410,56 @@ export function isBuildStepFunctionCall(step: BuildStepConfig): step is BuildSte
...
     }
   }
-  const calledFunctions = Array.from(calledFunctionsSet);
+  const calledFunctionsOrFunctionGroup = Array.from(calledFunctionsOrFunctionGroupsSet);
   const externalFunctionIdsSet = new Set(externalFunctionIds);
-  const nonExistentFunctions = calledFunctions.filter((calledFunction) => {
-    if (BuildFunction.isFulldIdNamespaced(calledFunction) && skipNamespacedFunctionsCheck) {
-      return false;
+  const externalFunctionGroupsIdsSet = new Set(externalFunctionGroupsIds);
+  const nonExistentFunctionsOrfunctionGroups = calledFunctionsOrFunctionGroup.filter(
```
sjchmiela, line 443:
```suggestion
  const nonExistentFunctionsOrFunctionGroups = calledFunctionsOrFunctionGroup.filter(
```

packages/build-tools/src/steps/utils/expoUpdates.ts
```diff
@@ -29,15 +29,23 @@ export async function configureEASUpdateIfInstalledAsync({
...
-    throw new Error(
-      `Cannot configure Expo Updates because the expo-updates package is not installed.`
-    );
+    if (inputs.throwIfNotConfigured) {
+      logger.error(
+        'Cannot configure EAS Update because the expo-updates package is not installed.'
+      );
+      throw new Error(
+        'Cannot configure EAS Update because the expo-updates package is not installed.'
+      );
```
sjchmiela, line 45: Why do we need to both log an error and throw an error? Shouldn't throwing be sufficient?

packages/build-tools/src/steps/utils/expoUpdates.ts
```diff
@@ -29,15 +29,23 @@ export async function configureEASUpdateIfInstalledAsync({
   inputs: {
     runtimeVersion?: string;
     channel?: string;
+    throwIfNotConfigured: boolean;
```
sjchmiela, line 32: nit: I was thinking this "mode" could be based off a parameter provided to `configureEASUpdateIfInstalledFunction`. Then the `fn` could even check `getExpoUpdatesPackageVersionIfInstalledAsync` before calling (or not calling) `configureEASUpdateIfInstalledAsync`. This would keep the oddity of `throwIfNotConfigured` in the `BuildFunction`, not leaking it to utils.

sjchmiela (approved):

PR #334, opened 2024-02-12
Put temporary directories in `os.tmpdir`

sjchmiela:
# Why

Turns out `fs.mkdtemp('install_maestro')` can [suddenly throw `EACCESS`](https://expo.dev/accounts/exponent/projects/eas-custom-builds-example/builds/59f6ccf3-a209-4c7d-b0dd-ab494fb62c04).

# How

Replaced `fs.mkdtemp('string')` with `fs.mkdtemp(path.join(os.tmpdir(), 'string'))`. This is what Node.js documentation suggests to do and also what we do in `www`.

# Test Plan

I verified this works right now on both [Android](https://expo.dev/accounts/exponent/projects/eas-custom-builds-example/builds/7ca68836-63ef-45bc-8ebf-207c827257e2) and [iOS](https://expo.dev/accounts/exponent/projects/eas-custom-builds-example/builds/97f592fa-24b9-4b73-bae4-f88379cf4b65) builds by copying `installMaestro` function into `my-custom-ts-function` and triggering builds manually. Both passed the "Install Maestro" step.

### DIFF
--- a/packages/build-tools/src/steps/functions/installMaestro.ts
+++ b/packages/build-tools/src/steps/functions/installMaestro.ts
@@ -1,7 +1,8 @@
 import assert from 'assert';
+import fs from 'fs';
+import os from 'os';
 import path from 'path';
 
-import fs from 'fs-extra';
 import {
   BuildFunction,
   BuildRuntimePlatform,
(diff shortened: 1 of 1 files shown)


packages/build-tools/src/steps/functions/installMaestro.ts
```diff
@@ -1,7 +1,8 @@
 import assert from 'assert';
+import fs from 'fs';
```
szdziedzic, line 2: I think we usually use `fs-extra` in other parts of the code. Maybe it's worth using it here as well?

szdziedzic (approved):

packages/build-tools/src/steps/functions/installMaestro.ts
sjchmiela: So in fact I moved this away from using `fs-extra`. If we can not use it, why use it?

szdziedzic: Oh, sorry, I see 🙈

sjchmiela: I mean, this may be a wrong decision and maybe we should decide to stick to `fs-extra`. I'm not sure though — for instance `www` only uses `fs`. Wanted to experiment and see if it works!

PR #335, opened 2024-02-13
[eas-build-job] [ENG-11196] Add optional buildExpoUrl field to Job

radoslawkrzemien:
# Why

We want to allow the user to customise their Slack messages in custom builds with build details. One of those should be the link to the build, which we don't want to be build step specific, since the same url will be valid and constant for all build steps. In such a case, the build link has been added as a new field to the Job object, and will be set either by cli or www before the build starts

# How

Added new field to the Job object and to schema validator

# Test Plan

Updated automated tests

### DIFF
--- a/packages/eas-build-job/src/__tests__/android.test.ts
+++ b/packages/eas-build-job/src/__tests__/android.test.ts
@@ -44,7 +44,7 @@ describe('Android.JobSchema', () => {
           SOME_ENV: '123',
         },
       },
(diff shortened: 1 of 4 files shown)


szdziedzic (approved):
lgtm

sjchmiela (approved):
`expoBuildUrl` reads better to me (as would Android artifact path), apart from that looks good.

PR #336, opened 2024-02-14
Add type to `staticContext`

sjchmiela:
# Why

We want to let users depend on `staticContext` properties.

# How

Added `BuildStaticContext` type and used it anywhere we reference `staticContext`.

I considered making everything generic with `TStaticContext` to better handle tests, but in the end decided to cast in tests `as unknown as BuildStaticContext`.

# Test Plan

This mostly changes types. Automated tests.


szdziedzic (approved):

szdziedzic:
Fix warning and we should be good to go

PR #337, opened 2024-02-15
Support glob in artifact path(s)

sjchmiela:
# Why

I was trying to add an example for uploading multiple artifacts — an IPA and an APP. I wanted to use `upload-artifact` to grab the APP, so I tried using one of https://github.com/expo/eas-build/blob/61fcb7d3662c8964da28272ca6c9af5626cb42fb/packages/build-tools/src/steps/functions/findAndUploadBuildArtifacts.ts#L96-L103 as path. As glob was not supported, this was not possible.

# How

As we call `findAndUploadArtifacts` before handing artifact off to `ctx.runtimeApi.uploadArtifact`, we'll do the same in `uploadArtifacts`.

Adds support for `paths` input which is expected to be an array of strings (as validated by Zod).

# Test Plan

Added an automated test.


szdziedzic (approved):
Did you test it e2e locally? Can you add an example of how a user should call this function? Should it be sth like:
```yml
eas/upload-artifact:
  type: sometype
  path: |
     glob1
     glob2
```
?

What will happen if we have 5 paths found for the app archive?

These questions are non-blocking ones, I'm just curious if I fully understand everything

sjchmiela:
> Did you test it e2e locally?

No, I trusted the integration test, but good call — I will test this e2e locally before merging.

> Can you add an example of how a user should call this function? (…)

Yup. https://github.com/expo/expo/pull/27122 Will add a comment here too. :+1:

> What will happen if we have 5 paths found for the app archive?

Good question. `findAndUploadArtifacts` would upload all of them, so theoretically this change makes `upload-artifact` more like `find-and-upload-artifacts` (which I find a positive thing). Question how does this tie with e.g. EAS Submit remains, unfortunately, but I think it is out of the scope of this pull request?

 https://github.com/expo/eas-build/blob/61fcb7d3662c8964da28272ca6c9af5626cb42fb/packages/build-tools/src/steps/functions/findAndUploadBuildArtifacts.ts#L17-L30

sjchmiela:
Zrzut ekranu 2024-02-19 o 22 24 51

packages/steps/package.json
```diff
@@ -48,6 +48,7 @@
     "node": ">=16.0.0"
   },
   "dependencies": {
+    "@expo/eas-build-job": "^1.0.71",
```
szdziedzic, line 51: I think that we should do the exact version as for other places
```suggestion
    "@expo/eas-build-job": "1.0.71",
```
so `lerna` can auto bump it when publishing a new release of `@expo/eas-build-job`.

BTW I don't see any changes to the `steps` package except this `package.json` change. Why is it needed?

sjchmiela: 👍 I didn't realize I committed that.

This is necessary so that users don't need to include both steps and eas-build-job in their custom functions to get staticContext.job typing, which only exists in eas-build-job.

PR #338, opened 2024-02-15
Add `eas/maestro_test` function group

sjchmiela:
# Why

We want to help developers get started with EAS Build + Maestro.

# How

Added another function group that takes care of setting up environment for Maestro, running the actual tests and uploading test results later.

Ideas for improvement:
- support multiple flows (implemented)
- skip setup of (Maestro, Simulator, Emulator) if a matching # is ready (not implemented)

# Test Plan

I ran it on my computer and:
- iOS Simulator could not boot
- Android script could not run:

Zrzut ekranu 2024-02-19 o 23 21 46

Maybe tomorrow it will work.

### DIFF
--- a/packages/build-tools/src/steps/functionGroups/maestroTest.ts
+++ b/packages/build-tools/src/steps/functionGroups/maestroTest.ts
@@ -1,0 +1,69 @@
+import {
+  BuildFunctionGroup,
+  BuildStep,
+  BuildStepInput,
+  BuildStepInputValueTypeName,
+} from '@expo/steps';
+import { Platform } from '@expo/eas-build-job';
+
+import { CustomBuildContext } from '../../customBuildContext';
+import { createInstallMaestroBuildFunction } from '../functions/installMaestro';
+import { createStartIosSimulatorBuildFunction } from '../functions/startIosSimulator';
+import { createStartAndroidEmulatorBuildFunction } from '../functions/startAndroidEmulator';
+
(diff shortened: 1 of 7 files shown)


packages/build-tools/src/steps/functionGroups/maestroTest.ts
```diff
@@ -0,0 +1,93 @@
...
+            id: BuildStep.getNewId(),
+            name: 'maestro_test',
+            displayName: 'Run "maestro test"',
+            ifCondition: '${ never() }',
+            fn: (ctx) => {
+              ctx.logger.warn('flow_file_path not provided. Skipping running test.');
+            },
+          })
+        );
+      }
```
szdziedzic, line 73: Maybe we should throw an error here? Not all of the users look at our website to figure out that the build is successful and in this case, you wouldn't even know that tests weren't run

szdziedzic (approved):
Looks good

packages/build-tools/src/steps/functionGroups/maestroTest.ts
sjchmiela: Right, initially I was planning to accept `maestro_test` with a single path argument (for single flow) or no path argument (and then user would add individual `maestro test flow/...` run steps manually). Then I figured we can start accepting multiple paths in more steps and forgot to adjust this. Will make `path` / `flow_file_path` required.

PR #340, opened 2024-02-21
[ENG-11387][eas-build-job] add submission log phases

szdziedzic:
# Why

We want to display logs better. To do so we need to display separate phases.

# How

Add phases for submission logs

### DIFF
--- a/packages/eas-build-job/src/logs.ts
+++ b/packages/eas-build-job/src/logs.ts
@@ -54,6 +54,13 @@ export enum BuildPhase {
   CUSTOM = 'CUSTOM',
 }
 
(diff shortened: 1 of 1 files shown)


sjchmiela (commented):
Why do we need a phase enum? Could phases be "named", not "keyed"?

sjchmiela (approved):
Approving to move forward. I think the way we do logs could be simplified by not requiring us to maintain enum and mapping in a third-party (from `www` and `turtle-v2`s point of view) library and instead use phase names directly. (`runPhase(fn, { name: "Submit to Play Store" })`). But that's out of the scope of current project.

PR #342, opened 2024-02-21
Log installed apps in `maestro_test`

sjchmiela:
# Why

I didn't like:
- "Install app to Simulator" printed out no logs
- Install Maestro printed out too many logs.

# How

For installing apps changed `command` into an `fn` which lets us use `fast-glob`, `logger` and in general feels simpler to maintain.

For hiding Maestro logs removed `logger` from `spawnAsync`.

Also added `HOMEBREW_NO_INSTALL_CLEANUP` so it doesn't clean up packages. Reduces install time by 3 seconds.

# Test Plan

Will deploy and test.

### DIFF
--- a/packages/build-tools/src/steps/functionGroups/maestroTest.ts
+++ b/packages/build-tools/src/steps/functionGroups/maestroTest.ts
@@ -3,8 +3,10 @@ import {
   BuildStep,
   BuildStepInput,
   BuildStepInputValueTypeName,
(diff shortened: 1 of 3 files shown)


sjchmiela:
Actually, I should revert the `command` -> `fn` change so the step is copiable by end users more easily. Let me do that.

szdziedzic (approved):

PR #346, opened 2024-02-23
[build-tools] Add outputs

radoslawkrzemien:
# Why

Added outputs to runGradle and runFastlane steps to allow sending them via Slack, for example

# How

Added output fields to `runGradle` and `runFastlane` steps. Now the user can use values like `steps.run_gradle.status_text` to see if the build succeeded or not, and if it fails they can use `steps.run_gradle.error_text` to get the error message.
These can then be used for example in `eas/send_slack_message` function to send a notification to the user's Slack channel

# Test Plan

Automatic tests pass

Tested manually
Config
Screenshot 2024-02-23 at 14 23 37
Messages
Screenshot 2024-02-23 at 14 24 23

### DIFF
--- a/packages/build-tools/src/steps/functionGroups/build.ts
+++ b/packages/build-tools/src/steps/functionGroups/build.ts
@@ -73,6 +73,7 @@ function createStepsForIosSimulatorBuild({
         throw_if_not_configured: false,
       },
     });
+  const runFastlane = runFastlaneFunction();
   return [
     createCheckoutBuildFunction().createBuildStepFromFunctionCall(globalCtx),
     createInstallNodeModulesBuildFunction().createBuildStepFromFunctionCall(globalCtx),
@@ -80,7 +81,7 @@ function createStepsForIosSimulatorBuild({
     installPods,
     configureEASUpdate,
     generateGymfileFromTemplateFunction().createBuildStepFromFunctionCall(globalCtx),
-    runFastlaneFunction().createBuildStepFromFunctionCall(globalCtx),
+    runFastlane.createBuildStepFromFunctionCall(globalCtx, { id: runFastlane.id }),
     createFindAndUploadBuildArtifactsBuildFunction(
       buildToolsContext
     ).createBuildStepFromFunctionCall(globalCtx),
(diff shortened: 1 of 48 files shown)


packages/build-tools/src/steps/functionGroups/build.ts
```diff
@@ -130,7 +132,7 @@ function createStepsForIosBuildWithCredentials({
     configureIosCredentialsFunction().createBuildStepFromFunctionCall(globalCtx),
     configureIosVersionFunction().createBuildStepFromFunctionCall(globalCtx),
     generateGymfile,
-    runFastlaneFunction().createBuildStepFromFunctionCall(globalCtx),
+    runFastlane.createBuildStepFromFunctionCall(globalCtx, { id: runFastlane.id }),
```
szdziedzic, line 135: What will be the `runFastlane.id` resolved to? If it's some nondeterministic random UUID I suggest making it const so users can reuse it.
```suggestion
    runFastlaneFunction().createBuildStepFromFunctionCall(globalCtx, { id: 'run_fastlane' }),
```

yarn.lock
```diff
@@ -673,6 +673,14 @@
...
     sucrase "3.34.0"
 
+"@expo/eas-build-job@1.0.74":
+  version "1.0.74"
+  resolved "https://registry.yarnpkg.com/@expo/eas-build-job/-/eas-build-job-1.0.74.tgz#041a9472f1e7505b53947db79f9a3e42e4384c96"
+  integrity sha512-8HfL8erYrZ5z38E+OqsS5xt2G5Ajt0eF/9RwcgtTGd0tNBn8Lmk8ICeazav2B3ETyK9SzJmRE2LtG+DTWdopvA==
+  dependencies:
+    joi "^17.11.0"
+    semver "^7.5.4"
+
```
szdziedzic, line 683: What triggered this install 🤔?

packages/build-tools/src/steps/functionGroups/build.ts
radoslawkrzemien: That's exactly the reason for this change - it was a nondeterministic random UUID, now it is the same as the id defined in the `steps/functions/runFastlane.ts`, which is `run_fastlane`

yarn.lock
radoslawkrzemien: `"@expo/eas-build-job": "1.0.74"` in `package.json

szdziedzic: Can we revert it?

radoslawkrzemien: Yeah, sure

radoslawkrzemien: Reverted

PR #347, opened 2024-02-26
[eas-build-job] allow secrets to be empty strings

szdziedzic:
# Why

https://exponent-internal.slack.com/archives/CLQ58M23X/p1708680627608489

# How

I believe `''` are valid env vars and we should allow it 🤔


kadikraman (approved):
It's surprising that an empty string is not a "string" in a Joi schema. So easy to miss this 🤯

sjchmiela (approved):

szdziedzic:
>It's surprising that an empty string is not a "string" in a Joi schema. So easy to miss this 🤯

yeah, true 😅

PR #351, opened 2024-02-27
[build-tools] [ENG-11196] Support object input for Block Kit messages

radoslawkrzemien:
# Why

Added support for object payload of slack messages for usage with the slack block kit
Depends on: https://github.com/expo/eas-build/pull/350

# How

Added a second input for the message contents - `payload`. It is mutually exclusive with `message` input, so providing both will result in an error, but one of them is required so providing none will also result in an error.
`payload` is a JSON type input and contains the Slack message in their Block Kit layout form.
With the recent changes in the `steps` package it is also possible to interpolate values in these message blocks for richer messages with build details

# Test Plan

Automatic tests pass
Tested manually
Config
Screenshot 2024-02-27 at 19 27 50
Screenshot 2024-02-27 at 19 27 56
Screenshot 2024-02-27 at 19 28 02
Message
Screenshot 2024-02-27 at 19 21 53

### DIFF
--- a/packages/build-tools/src/steps/functions/sendSlackMessage.ts
+++ b/packages/build-tools/src/steps/functions/sendSlackMessage.ts
@@ -29,17 +29,11 @@ export function createSendSlackMessageFunction(): BuildFunction {
       const slackMessage = inputs.message.value as string;
       const slackPayload = inputs.payload.value as object;
       if (!slackMessage && !slackPayload) {
-        logger.warn(
-          'You need to provide either "message" input or "payload" input to specify the Slack message contents'
-        );
         throw new Error(
           'You need to provide either "message" input or "payload" input to specify the Slack message contents'
         );
       }
       if (slackMessage && slackPayload) {
-        logger.warn(
-          'You cannot specify both "message" input and "payload" input - choose one for the Slack message contents'
-        );
         throw new Error(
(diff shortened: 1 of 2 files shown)


packages/build-tools/src/steps/functions/sendSlackMessage.ts
```diff
@@ -22,6 +27,23 @@ export function createSendSlackMessageFunction(): BuildFunction {
...
       const { logger } = stepCtx;
       const slackMessage = inputs.message.value as string;
+      const slackPayload = inputs.payload.value as object;
+      if (!slackMessage && !slackPayload) {
+        logger.warn(
+          'You need to provide either "message" input or "payload" input to specify the Slack message contents'
+        );
+        throw new Error(
+          'You need to provide either "message" input or "payload" input to specify the Slack message contents'
+        );
```
szdziedzic, line 37: Is it necessary to have both of them?

szdziedzic (approved):

packages/build-tools/src/steps/functions/sendSlackMessage.ts
radoslawkrzemien: Probably not, since the error is logged anyway

PR #354, opened 2024-03-01
[build-tools] Update expo-updates runtime version resolution

wschurman:
# Why

Same as https://github.com/expo/eas-cli/pull/2251 but for build-tools. This issue (using a single version of config plugins to execute updates related logic rather than the version of config plugins the app uses) seems to be pretty widespread.

This PR applies the same strategy: call into the app versioned expo-updates CLI to resolve runtime version (if possible, falling back to single version config plugins for now).

# How

Copy paste code.

# Test Plan

1. Follow instructions to do build locally using local copy of eas-cli and eas-build.
2. run `~/expo/run-with-local-eas-build.sh build --local`, see runtime version is consistently resolved.


wschurman:
Ping

quinlanj (approved):

PR #356, opened 2024-03-06
Revert "Revert "[build-tools] Update expo-updates runtime version resolution (#354)" (#355)"

wschurman:
# Why

This reverts commit https://github.com/expo/eas-build/commit/39e063ac5cb25db8809d86110067b3608911f3f3.

Then, this applies a fix on top of it to support older SDKs.

# How

Revert, re-test with SDK 45-50.

# Test Plan

```
yarn create expo-app kjdasjkds555hjkdashjk50 --template blank@50
eas init
eas update:configure
~/expo/run-with-local-eas-build.sh build --local
```

(the last step uses a local eas-cli to patch the revert in that repo: https://github.com/expo/eas-cli/pull/2265)

### DIFF
--- a/packages/build-tools/src/utils/expoUpdatesCli.ts
+++ b/packages/build-tools/src/utils/expoUpdatesCli.ts
@@ -22,8 +22,11 @@ export async function expoUpdatesCommandAsync(projectDir: string, args: string[]
   try {
     return (await spawnAsync(expoUpdatesCli, args)).stdout;
   } catch (e: any) {
-    if (e.stderr) {
-      if ((e.stderr as string).includes('Invalid command')) {
+    if (e.stderr && typeof e.stderr === 'string') {
+      if (
+        e.stderr.includes('Invalid command') || // SDK 51+
+        e.stderr.includes('commands[command] is not a function') // SDK <= 50
+      ) {
         throw new ExpoUpdatesCLIInvalidCommandError(
           `The command specified by ${args} was not valid in the \`expo-updates\` CLI.`
         );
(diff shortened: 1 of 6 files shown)


quinlanj (approved):

packages/build-tools/src/utils/expoUpdatesCli.ts
```diff
@@ -22,8 +22,11 @@ export async function expoUpdatesCommandAsync(projectDir: string, args: string[]
   try {
     return (await spawnAsync(expoUpdatesCli, args)).stdout;
   } catch (e: any) {
-    if (e.stderr) {
-      if ((e.stderr as string).includes('Invalid command')) {
+    if (e.stderr && typeof e.stderr === 'string') {
+      if (
+        e.stderr.includes('Invalid command') || // SDK 51+
+        e.stderr.includes('commands[command] is not a function') // SDK <= 50
+      ) {
```
quinlanj, line 29: Is it possible to determine the expo SDK we are dealing with? If we support many many old SDK's, perhaps something like `resolveFrom('expo/package.json')` and getting the version from there would be more robust

wschurman: We'd need to resolve the expo-updates package version (which we do already in some places) but I'd like to move away from if possible since things like canary versions don't work very well with it (background on that decision: https://github.com/expo/expo/pull/27143).

For that reason I prefer just trying/catching, but I could be swayed the other way.

wschurman:
I need to figure out how to add tests for this, but it's probably outside the scope of this PR since they'd need to be e2e tests to catch this sort of issue.

quinlanj (approved):

PR #357, opened 2024-03-08
[ENG-11416][eas-build-job] add github build trigger options to job object

szdziedzic:
# Why

Companion to https://github.com/expo/eas-cli/pull/2271

We need this information so we know whether to use the `--auto-submit` flag in Turtle worker when running `eas build:internal`. This information should be provided by WWW.

# How

Add `githubTriggerOptions` to the job object

# Test Plan

Tests

### DIFF
--- a/packages/eas-build-job/src/android.ts
+++ b/packages/eas-build-job/src/android.ts
@@ -107,6 +107,10 @@ export interface Job {
     prebuildCommand?: string;
   };
   expoBuildUrl?: string;
+  gitHubTriggerOptions?: {
+    autoSubmit: boolean;
+    submitProfile?: string;
+  };
 }
 
 const SecretsSchema = Joi.object({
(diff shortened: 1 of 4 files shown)


packages/eas-build-job/src/ios.ts
```diff
@@ -121,6 +121,10 @@ export interface Job {
     prebuildCommand?: string;
   };
   expoBuildUrl?: string;
+  gitHubTriggerOptions?: {
```
sjchmiela, line 124:
```suggestion
  githubTriggerOptions?: {
```

packages/eas-build-job/src/android.ts
```diff
@@ -107,6 +107,10 @@ export interface Job {
     prebuildCommand?: string;
   };
   expoBuildUrl?: string;
+  gitHubTriggerOptions?: {
```
sjchmiela, line 110:
```suggestion
  githubTriggerOptions?: {
```

And this is why I love Zod.

sjchmiela (approved):

PR #360, opened 2024-03-11
[build-tools] Upload Xcode logs for custom builds behind the scenes

sjchmiela:
# Why

They're good to have to debug things. https://exponent-internal.slack.com/archives/C04GPRQCEBG/p1709836166859129

# How

Added fail-safe uploading Xcode logs behind the scenes to custom builds. Also refactored `findAndUploadArtifacts` a bit. One notable change is changing `cwd` for uploading build artifacts from `projectTargetDirectory` to `workingDirectory`. This is what `applicationArchivePath` uses. Which one should it be? Why?

# Test Plan

Added some tests.

### DIFF
--- a/packages/build-tools/src/__mocks__/fs.ts
+++ b/packages/build-tools/src/__mocks__/fs.ts
@@ -1,6 +1,6 @@
 import { fs } from 'memfs';
 
-fs.mkdirSync('/tmp');
+fs.mkdirSync('/tmp', { recursive: true });
 if (process.env.TMPDIR) {
   fs.mkdirSync(process.env.TMPDIR, { recursive: true });
 }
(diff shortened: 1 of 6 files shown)


szdziedzic (approved):
>One notable change is changing cwd for uploading build artifacts from projectTargetDirectory to workingDirectory. This is what applicationArchivePath uses. Which one should it be? Why?

I think It was the `projectTargetDirectory` (which I think is a default `workingDir`) because we don't allow users to enter their globs to search for build artifacts as input and our ones are working only for default working dir (react native project dir) if I'm not wrong.

szdziedzic:
Do we still want to switch to using working dir?

sjchmiela:
> Do we still want to switch to using working dir?

Unless I'm missing something, there's little reason to have different `cwd`s for application archive and for build artifacts. When defining those properties in `eas.json`, developers expected them to be resolved from the same directory and so they should always be.

If we agree on this, we should use either working directory for both or `projectTargetDirectory` for both.

I think if we allow changing `working_directory` for any step, the change should be applied to the step. Hence my suggestion to use working directory as `cwd`.

What do you think?

PR #362, opened 2024-03-13
[build-tools] custom builds equivalent of move runtime version calculation ahead of mutations

szdziedzic:
# Why

Custom builds equivalent of https://github.com/expo/eas-build/pull/361

# How

Create a new function `eas/calculate_eas_update_runtime_version`. Use the output of this function as an input to `eas/configure_eas_update`, `eas/run_gradle`, and `eas/run_fastlane` functions.

Apply changes to `eas/build`.

Refactor `eas/configure_eas_update` to do the same as its equivalent in the standard build process.

# Test Plan

Tests, run builds locally


wschurman (approved):

wschurman:
Thanks!

PR #364, opened 2024-03-13
Add `eas/resolve_build_config` function

sjchmiela:
# Why

GitHub build job data is missing a lot of properties, because we don't provide them when we create the build, because we can't, because the config may need to be resolved dynamically, with external dependencies etc.

For this reason `eas build:internal` exists which we call during the build, once all dependencies are installed. We set this up in `get_credentials_for_build_triggered_by_github_integration`. This covered half of the use cases. Turns out we need to call this also in builds that don't need credentials so they have `buildType`, `version` and other properties.

# How

Added `eas/resolve_build_config` function that will be called in _all_ `eas/build` flows. If the build has been triggered by a GitHub trigger, it will call `eas build:internal` to resolve necessary properties, update build information in `www` and in context. It will always print the configuration (with `secrets` and `projectArchive` omitted, as we omit it in the regular build process).

I'm not sure if printing all or even some job properties makes sense. For me, the advantage is that a user can inspect what is the configuration, what properties can they access in `${ eas.job. ... }`. On the other hand it feels noisy. What is your opinion?

# Test Plan

Zrzut ekranu 2024-03-13 o 22 22 41

### DIFF
--- a/packages/build-tools/src/steps/easFunctions.ts
+++ b/packages/build-tools/src/steps/easFunctions.ts
@@ -23,6 +23,7 @@ import { createInstallMaestroBuildFunction } from './functions/installMaestro';
 import { createGetCredentialsForBuildTriggeredByGithubIntegration } from './functions/getCredentialsForBuildTriggeredByGitHubIntegration';
 import { createInstallPodsBuildFunction } from './functions/installPods';
 import { createSendSlackMessageFunction } from './functions/sendSlackMessage';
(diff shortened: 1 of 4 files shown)


szdziedzic (approved):
I think it can be generally a cool thing for users to be able to inspect job contents. What I'm worrying about is the situation when some user decides that this is too noisy for them and they don't want all of this stuff printed and they stop using this step. But I think it's unlikely.

PR #367, opened 2024-03-18
[eas-build-job] Add generic jobs schema

sjchmiela:
# Why

We need types for generic jobs. They're slimmer than builds'.

# How

Added `zod` for validation. Added slimmed down types.

# Test Plan

Currently developing alongside `turtle-v2` changes.


packages/eas-build-job/src/common.ts
```diff
@@ -82,6 +83,27 @@ export const ArchiveSourceSchema = Joi.object<ArchiveSource>({
     }),
   });
 
+export const ArchiveSourceSchemaZ = z.discriminatedUnion('type', [
+  z.object({
+    type: z.literal(ArchiveSourceType.GIT),
+    repositoryUrl: z.string(),
```
szdziedzic, line 89: maybe 
```suggestion
    repositoryUrl: z.string().url(),
```
?

szdziedzic (approved):
looks good, if we missed something we can always come back and add it because it won't be user-facing for a while

PR #368, opened 2024-03-19
[build-tools] run `resolveRuntimeVersionForExpoUpdatesIfConfiguredAsync` only when `job.updates.channel` is set

szdziedzic:
# Why

We run other EAS Update configuration steps only when `job.updates.channel` is set (the user set the `channel` in their `eas.json`) and [EAS Update is configured](https://github.com/expo/eas-build/blob/0a3819f36c1244a859c731571116c2f9e6c19112/packages/build-tools/src/utils/expoUpdates.ts#L276-L295):

https://github.com/expo/eas-build/blob/0a3819f36c1244a859c731571116c2f9e6c19112/packages/build-tools/src/utils/expoUpdates.ts#L165-L173

# How

Don't run `resolveRuntimeVersionForExpoUpdatesIfConfiguredAsync` if `channel` is falsy or EAS Update is not configured (`appconfig.url` is invalid or falsy)

# Test Plan

Tests

### DIFF
--- a/packages/build-tools/src/builders/android.ts
+++ b/packages/build-tools/src/builders/android.ts
@@ -77,6 +77,7 @@ async function buildAsync(ctx: BuildContext<Android.Job>): Promise<void> {
         logger: ctx.logger,
         appConfig: ctx.appConfig,
         platform: ctx.job.platform,
+        channel: ctx.job.updates?.channel,
       });
     }
   );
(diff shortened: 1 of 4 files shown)


wschurman (changes requested):
I think we want people to be able to use expo-updates (and the fingerprint runtime version policy as well as other policies) with EAS Build even if they don't use EAS Update. Put another way, if they point expo-updates to their own update backend, they should still be able to use EAS Build.

`channel` is a EAS-Update-only concept and is not required for non EAS Update usage of expo-updates.

wschurman:
IMO it's fine to run this step whenever `expo-updates` (the package) is installed.

PR #371, opened 2024-03-26
Expand Job definition

sjchmiela:
# Why

I'm thinking we should have `Job = Android.Job | Ios.Job | Generic.Job` so we can use the same `Job` concept in existing functions (some of which will work for generic jobs too) and worker code.

# How

This, however, as one can see, requires a lot of, mostly, typing changes.

# Test Plan

TSC passes. Tests should too.


packages/eas-build-job/src/generic.ts
```diff
@@ -31,5 +31,8 @@ export namespace Generic {
     }),
     expoDevUrl: z.string().url(),
     builderEnvironment: BuilderEnvironmentSchemaZ,
+    // We use this to discern between Android.Job, Ios.Job and Generic.Job.
+    platform: z.never().optional(),
+    triggeredBy: z.literal(BuildTrigger.GIT_BASED_INTEGRATION),
```
szdziedzic, line 36: Generic jobs can be triggered only from GH?

packages/eas-build-job/src/index.ts
```diff
@@ -13,10 +16,12 @@ export {
...
-export { Job, sanitizeJob } from './job';
+export { BuildJob, sanitizeBuildJob } from './job';
 export { Metadata, sanitizeMetadata } from './metadata';
 export * from './logs';
 export * as errors from './errors';
 export * from './artifacts';
 export * from './context';
 export * from './generic';
+
+export type Job = BuildJob | Generic.Job;
```
szdziedzic, line 27: Can we put it in `job.ts`? I mean this way we won't need to import it from `index.ts` in the files that are imported in the `index.ts` themselves.

packages/eas-build-job/src/generic.ts
```diff
@@ -31,5 +31,8 @@ export namespace Generic {
     }),
     expoDevUrl: z.string().url(),
     builderEnvironment: BuilderEnvironmentSchemaZ,
+    // We use this to discern between Android.Job, Ios.Job and Generic.Job.
```
szdziedzic, line 34: We have a `type` field in the `Job` object (I mean `BuildJob`). Maybe we can use this field somehow to distinguish between generic jobs and build jobs?
https://github.com/expo/eas-build/blob/a8c2f0d7b24defc7016987a3d7f400f7808f4a29/packages/eas-build-job/src/android.ts#L69

packages/eas-build-job/src/job.ts
```diff
@@ -4,17 +4,17 @@ import { Platform } from './common';
...
+export type BuildJob = Android.Job | Ios.Job;
 
-export const JobSchema = Joi.object<Job>({
+export const JobSchema = Joi.object<BuildJob>({
   platform: Joi.string()
     .valid(...Object.values(Platform))
     .required(),
 })
   .when(Joi.object({ platform: Platform.ANDROID }).unknown(), { then: Android.JobSchema })
   .when(Joi.object({ platform: Platform.IOS }).unknown(), { then: Ios.JobSchema });
```
szdziedzic, line 15: Maybe if we distinguish between `BuildJob` and `GenericJob` based on the `type` field we can have one conditional Joi/Zed schema to validate Jobs? I'm not sure if there is any benefit to it. Just thinking out loud.

packages/build-tools/src/context.ts
```diff
@@ -282,6 +282,11 @@ export class BuildContext<TJob extends Job> {
   }
 
   public getReactNativeProjectDirectory(baseDirectory = this.buildDirectory): string {
+    if (!this.job.platform) {
+      // Generic jobs start from base directory.
+      return baseDirectory;
+    }
```
szdziedzic, line 288: I know it's all tricky, but do we even need to use BuildContext for generic jobs? I mean it seems like a build job only concept.

packages/build-tools/src/steps/easFunctions.ts
```diff
@@ -46,10 +46,22 @@ export function getEasFunctions(ctx: CustomBuildContext): BuildFunction[] {
...
+
+  if (ctx.hasBuildJob()) {
+    functions.push(
+      ...[
+        createFindAndUploadBuildArtifactsBuildFunction(ctx),
+        createResolveBuildConfigBuildFunction(ctx),
+        createGetCredentialsForBuildTriggeredByGithubIntegration(ctx),
+      ]
+    );
+  }
```
szdziedzic, line 64: Why did you pick these as BuildJob only and not the others? I mean I'm not saying it's wrong, just want to understand it better

szdziedzic (commented):
Some questions:

packages/eas-build-job/src/job.ts
sjchmiela: I would love to have one Zod schema to validate everything! I didn't want to embark on a journey to rewrite all (or at least Job-related) Joi schemas to Zod. If I remember correctly it touches `www` too and is in general a bigger project to do.

packages/build-tools/src/steps/easFunctions.ts
sjchmiela: Of course! What I did is change the meaning of `Job` to include `Generic.Job`. This, obviously caused some TS errors, like "Property `type` does not exist [in `job`]". Then, I inspected each error and:
- if a function made sense in Generic Jobs context, I made it support it (by adding `'type' in job` and others)
- if a function did not make sense in Generic Jobs context (like these), I changed the expected context type to specify `BuildJob`. Then I couldn't pass a `Job` context to these functions, so I excluded them like so.

So, essentially, I'm excluding only the functions that don't work for `Generic.Job`. Maybe more functions don't make sense, here I'm only focusing on type layer, not "product layer".

packages/build-tools/src/context.ts
sjchmiela: Hmm, so I think a lot of `BuildContext` is reusable for generic jobs too. All build phases logic, env, artifacts…

I agree some parts are build-specific, like `getReactNativeProjectDirectory`, `appConfig`.

Maybe what we'll need to do is have a `TaskContext<TJob>` which would have build phases, env, artifacts and have a `BuildContext < TaskContext<BuildJob>` with `appConfig` and stuff. I didn't want to do too much in one pull request though, so my assumption for now is that `BuildContext` is like `TaskContext`.

packages/eas-build-job/src/index.ts
sjchmiela: Moved. 👍

packages/eas-build-job/src/generic.ts
sjchmiela: This is a good suggestion – I'm not a huge fan of using `platform` as the discriminator either.

However, I wouldn't want to extend the `Workflow` enum to accommodate generic jobs either (funnily enough, there already is a `Workflow.GENERIC` which means a completely different thing.

We could add a `jobType: "build" | "jobRun"` property, but:
- we'll need to support `jobType: undefined` nonetheless (for backwards compatibility)
- having `platform: undefined` actually has a lot of good consequences, because we use `platform` for many things, so we get to recheck all those for generic jobs.

Given this, what is your thinking?

szdziedzic: >However, I wouldn't want to extend the Workflow enum to accommodate generic jobs either (funnily enough, there already is a Workflow.GENERIC which means a completely different thing.

Oh, I forgot about it 😅 

>we'll need to support `jobType: undefined` nonetheless (for backwards compatibility)

Maybe we can make `the jobType` required and transform the `undefined` value to `build` by default using the `.default()` option in the Joi/Zod schema?

I mean having a different `jobType` seems like the cleanest approach to me, but it's not a string opinion.

I'm fine with proceeding as it is or adjusting it in a separate PR.

szdziedzic (approved):
It looks reasonable to me. I understand the change a way better with all of these answers, thanks!

packages/eas-build-job/src/generic.ts
sjchmiela: I'll proceed with the current state for now, who knows what else we'll do here haha.

PR #373, opened 2024-03-28
Export `runGenericJobAsync`

sjchmiela:
# Why

We'll use it to trigger generic jobs. For a second I thought we could reuse `runCustomBuildAsync`, but then I figured maybe it's better to have something separate.

Also, wraps preparing project with a build phase. We don't have this in custom build. I think it makes sense to track the time of archive / Git repository download? Or maybe not?

# How

Copied `runCustomBuildAsync`, removed a bunch of things. Reverted changes to `runCustomBuildAsync` done recently.

# Test Plan

This adds something new, so should be ok.

### DIFF
--- a/packages/build-tools/src/generic.ts
+++ b/packages/build-tools/src/generic.ts
@@ -13,9 +13,7 @@ import { getEasFunctionGroups } from './steps/easFunctionGroups';
 export async function runGenericJobAsync(ctx: BuildContext<Generic.Job>): Promise<void> {
   const customBuildCtx = new CustomBuildContext(ctx);
 
-  await ctx.runBuildPhase(BuildPhase.PREPARE_PROJECT, async () => {
-    await prepareProjectSourcesAsync(ctx, customBuildCtx.projectSourceDirectory);
-  });
+  await prepareProjectSourcesAsync(ctx, customBuildCtx.projectSourceDirectory);
 
   const relativeConfigPath = nullthrows(
     ctx.job.customBuildConfig?.path,
--- a/packages/build-tools/src/generic.ts
+++ b/packages/build-tools/src/generic.ts
@@ -1,0 +1,49 @@
+import path from 'path';
+
+import { BuildPhase, Generic } from '@expo/eas-build-job';
+import { BuildConfigParser, BuildStepGlobalContext, errors } from '@expo/steps';
+import nullthrows from 'nullthrows';
+
+import { BuildContext } from './context';
+import { prepareProjectSourcesAsync } from './common/projectSources';
+import { getEasFunctions } from './steps/easFunctions';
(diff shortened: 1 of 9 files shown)


packages/build-tools/src/generic.ts
```diff
@@ -0,0 +1,49 @@
...
+import { getEasFunctions } from './steps/easFunctions';
+import { CustomBuildContext } from './customBuildContext';
+import { getEasFunctionGroups } from './steps/easFunctionGroups';
+
+export async function runGenericJobAsync(ctx: BuildContext<Generic.Job>): Promise<void> {
+  const customBuildCtx = new CustomBuildContext(ctx);
+
+  await ctx.runBuildPhase(BuildPhase.PREPARE_PROJECT, async () => {
+    await prepareProjectSourcesAsync(ctx, customBuildCtx.projectSourceDirectory);
+  });
```
szdziedzic, line 18: To make it more interesting/interactive for a user it would be good if it printed some logs because this phase will be displayed on the website anyway, but as an empty phase. 

If we add logging to this function we should be aware that we probably would need to wrap it in a build phase for normal builds as well or only print logs for generic jobs.

szdziedzic: It can also be confusing for the users as normally the archive is downloaded during the `checkout` function run. Our checkout is weird and that could have been a reason why this was kind of hidden 🤔

packages/build-tools/src/generic.ts
```diff
@@ -0,0 +1,49 @@
...
+import { BuildConfigParser, BuildStepGlobalContext, errors } from '@expo/steps';
+import nullthrows from 'nullthrows';
+
+import { BuildContext } from './context';
+import { prepareProjectSourcesAsync } from './common/projectSources';
+import { getEasFunctions } from './steps/easFunctions';
+import { CustomBuildContext } from './customBuildContext';
+import { getEasFunctionGroups } from './steps/easFunctionGroups';
+
+export async function runGenericJobAsync(ctx: BuildContext<Generic.Job>): Promise<void> {
```
szdziedzic, line 13: Could you briefly sum up what's the difference between this function and `runCustomBuild`?

sjchmiela: - no `resolveEnvFromBuildProfileAsync` (even if we wanted to keep it, we don't have "job profiles" yet, so it'd be a todo)
- `configPath` doesn't use `getReactNativeDirectory`, only base
- it doesn't try to upload Xcode logs
- it doesn't return artifacts

packages/build-tools/src/generic.ts
sjchmiela: Hmm, sure. Removed wrapper.

PR #375, opened 2024-04-04
[eas-build-job] allow to set build logger level in job object

szdziedzic:
# Why

We currently set the logger level globally per `worker` service, so we never see debug logs in staging/production because the level there is just `info`. It might be helpful to allow setting logger level on a per-build basis.

I want to allow people to use it like `eas build ... --logger-level debug`

https://exponent-internal.slack.com/archives/C02123T524U/p1711730641664359

@wschurman highlighted a need for something like this

# How

Add `loggerLevel` to the job object

# Test Plan

Tests

### DIFF
--- a/packages/eas-build-job/package.json
+++ b/packages/eas-build-job/package.json
@@ -26,6 +26,7 @@
     "typescript": "^5.3.3"
   },
   "dependencies": {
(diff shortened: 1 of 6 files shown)


sjchmiela (approved):
Cool!

PR #383, opened 2024-04-22
[ENG-12057][build-tools] fail with meaningful error message if `ios.bundleIdentifier` or `android.package` is not set for GitHub builds

szdziedzic:
# Why

https://linear.app/expo/issue/ENG-12057/fail-with-a-nice-and-meaningful-error-message-if-iosbundleidentifier

# How

Fail with a meaningful error message if `ios.bundleIdentifier` or `android.package` is not set for GitHub builds

### DIFF
--- a/packages/build-tools/src/common/setup.ts
+++ b/packages/build-tools/src/common/setup.ts
@@ -67,17 +67,17 @@ export async function setupAsync<TJob extends BuildJob>(ctx: BuildContext<TJob>)
   });
 
   if (ctx.job.triggeredBy === BuildTrigger.GIT_BASED_INTEGRATION) {
-    if (!ctx.appConfig.ios?.bundleIdentifier) {
-      throw new Error(
-        'The "ios.bundleIdentifier" is required to be set in app config for builds triggered by GitHub integration. Learn more: https://docs.expo.dev/versions/latest/config/app/#bundleidentifier.'
-      );
-    }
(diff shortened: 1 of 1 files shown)


khamilowicz (approved):

packages/build-tools/src/common/setup.ts
```diff
@@ -59,7 +59,24 @@ export async function setupAsync<TJob extends BuildJob>(ctx: BuildContext<TJob>)
...
 
+  await ctx.runBuildPhase(BuildPhase.READ_APP_CONFIG, async () => {
+    const appConfig = ctx.appConfig;
+    ctx.logger.info('Using app configuration:');
+    ctx.logger.info(JSON.stringify(appConfig, null, 2));
+    await validateAppConfigAsync(ctx, appConfig);
+  });
+
   if (ctx.job.triggeredBy === BuildTrigger.GIT_BASED_INTEGRATION) {
+    if (!ctx.appConfig.ios?.bundleIdentifier) {
```
radoslawkrzemien, line 70: Can the GitHub build be triggered for only 1 platform? If so, wouldn't running an Android build always fail on this condition, and iOS build on the second?

szdziedzic: yeah, it can, makes sense

szdziedzic: fixed

PR #384, opened 2024-04-24
[build-tools] Pass through workflow from job to expo-updates CLI commands

wschurman:
# Why

`eas-build` change corresponding to https://github.com/expo/expo/pull/28403, which allows passing the workflow through from the job. This is necessary in EAS Build since it does prebuild, and workflow auto-detection in expo-updates depends on the presence of native directories or lack thereof, as well as their gitignore status.

Because our default gitignore doesn't ignore the native directories, prebuilt will make the detection code think it's a generic workflow.

# How

Pass workflow through. Note that in this library, there's a third workflow enum "UNKNOWN", which if passed through to the expo-updates CLIs will throw. This is on purpose as we shouldn't be running those CLIs for jobs where the workflow is unknown. Note that eas-cli only ever generates generic or managed when constructing a normal build job (the only type of job where these are run).

# Test Plan

Follow test plan of https://github.com/expo/expo/pull/28403 to create the project, then run the following to ensure runtime version resolution works with new arg:

`~/expo/run-with-local-eas-build.sh build --local`

### DIFF
--- a/packages/build-tools/src/builders/android.ts
+++ b/packages/build-tools/src/builders/android.ts
@@ -77,6 +77,7 @@ async function buildAsync(ctx: BuildContext<Android.Job>): Promise<void> {
         logger: ctx.logger,
         appConfig: ctx.appConfig,
         platform: ctx.job.platform,
+        workflow: ctx.job.type,
       });
     }
   );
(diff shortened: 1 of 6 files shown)


packages/build-tools/src/builders/android.ts
```diff
@@ -105,7 +106,12 @@ async function buildAsync(ctx: BuildContext<Android.Job>): Promise<void> {
       gradleCommand,
       androidDir: path.join(ctx.getReactNativeProjectDirectory(), 'android'),
       ...(resolvedExpoUpdatesRuntimeVersion
-        ? { extraEnv: { EXPO_UPDATES_FINGERPRINT_OVERRIDE: resolvedExpoUpdatesRuntimeVersion } }
+        ? {
+            extraEnv: {
+              EXPO_UPDATES_FINGERPRINT_OVERRIDE: resolvedExpoUpdatesRuntimeVersion,
+              EXPO_UPDATES_WORKFLOW_OVERRIDE: ctx.job.type,
```
wschurman, line 112: this isn't technically needed since this is only used if `EXPO_UPDATES_FINGERPRINT_OVERRIDE` isn't provided, but worth adding for consistency just in case someone puts up a project that uses it.

szdziedzic (approved):
looks good

PR #386, opened 2024-04-25
[build-tools] [ENG-11883] Add retries to artifact upload

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-11883/retry-uploading-artifacts-to-gcs-in-turtle-workers

# How

Wrapped the inner call to upload function within `BuildContext.uploadArtifact()` in a retry.
Set the options to try 2 more times with 30s interval.

# Test Plan

Added automated tests to check the behaviour

### DIFF
--- a/packages/build-tools/src/context.ts
+++ b/packages/build-tools/src/context.ts
@@ -2,28 +2,29 @@ import path from 'path';
 
 import fs from 'fs-extra';
 import {
-  ManagedArtifactType,
   BuildPhase,
   BuildPhaseResult,
   BuildPhaseStats,
-  Job,
-  LogMarker,
   Env,
-  errors,
-  Metadata,
   EnvironmentSecretType,
+  errors,
   GenericArtifactType,
   isGenericArtifact,
+  Job,
+  LogMarker,
+  ManagedArtifactType,
+  Metadata,
 } from '@expo/eas-build-job';
 import { ExpoConfig } from '@expo/config';
 import { bunyan } from '@expo/logger';
-import { SpawnPromise, SpawnOptions, SpawnResult } from '@expo/turtle-spawn';
+import { SpawnOptions, SpawnPromise, SpawnResult } from '@expo/turtle-spawn';
 import { BuildTrigger } from '@expo/eas-build-job/dist/common';
 
 import { PackageManager, resolvePackageManager } from './utils/packageManager';
 import { resolveBuildPhaseErrorAsync } from './buildErrors/detectError';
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/context.ts
```diff
@@ -226,7 +228,18 @@ export class BuildContext<TJob extends Job = Job> {
...
+        return await this._uploadArtifact({ artifact, logger });
+      },
+      {
+        retryOptions: {
+          retries: 2,
+          retryIntervalMs: this.ARTIFACT_UPLOAD_RETRY_INTERVAL_MS,
+        },
+        logger,
+      }
+    );
```
szdziedzic, line 242: I think we should only retry on DNS errors. When there’s a DNS error, we know that fetch() never connected to the destination server, so it’s always safe to retry. See how we do it in WWW: https://github.com/expo/universe/blob/8b03af4877629a5b0e6b694b241e187d7172da6e/server/www/src/utils/retryOnDNSFailure.ts

I believe the ones we've seen so far were always DNS ones:
image

https://exponent-internal.slack.com/archives/C031ZK9UWAC/p1713985105014779

szdziedzic: All of the `uploadArtifact` functions are created in `worker` service and I believe they use https://github.com/expo/turtle-v2/blob/09461a622f1f03b6c4f4dbc1c5d8b86831c4b0c5/src/libs/gcs/src/client.ts#L37-L57 under the hood, but please double check it with https://github.com/expo/turtle-v2/blob/09461a622f1f03b6c4f4dbc1c5d8b86831c4b0c5/src/services/worker/src/context.ts#L73-L119

radoslawkrzemien: Made the retry conditional on receiving DNS error, modelled after what is in place in www

radoslawkrzemien: From what I can see, all of the upload calls in the `uploadArtifact` function created in the `worker` do use `uploadWithSignedUrl()` under the hood

radoslawkrzemien: Hm, now I see that `uploadApplicationArchive` kinda uses it's own retry here - https://github.com/expo/turtle-v2/blob/main/src/services/worker/src/upload.ts#L96-L113
It isn't used by other artifact uploads, and it does not differentiate between different errors.
I presume this should be removed then, if we're adding a master retry to `build-tools`?

radoslawkrzemien: Unless there is a reason why this particular artifact upload already has it's own retry that I don't know about

PR #391, opened 2024-05-02
[build-tools][build-job] Remove classic updates concepts

wschurman:
# Why

`eas-build` counterpart to github.com/expo/eas-cli/pull/2357. Classic updates is no longer supported in any SDK version, so this code can be removed.

Closes ENG-12141.

# How

Remove code.

# Test Plan

`yarn tsc`

### DIFF
--- a/packages/build-tools/src/android/__tests__/expoUpdates.test.ts
+++ b/packages/build-tools/src/android/__tests__/expoUpdates.test.ts
@@ -5,10 +5,8 @@ import { AndroidConfig } from '@expo/config-plugins';
 
 import {
   AndroidMetadataName,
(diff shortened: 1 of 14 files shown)


quinlanj (approved):
LGTM. Putting in our slack discussion notes here for posterity:

TLDR; new builds will never receive old Classic update publishes so it makes sense to remove the code path altogether
- We are removing code from EAS Build so folks can no longer make classic update builds (even if they are on a prior SDK)
- We announced we would no longer accepting Classic Update publishes, but would still be serving updates.
- In theory someone could have old publishes and could make a build to support it. 
- The embedded update in the build would be newer than the old publish, so we would never end up serving the update to the build

PR #392, opened 2024-05-03
[build-tools] Set up a branch or a tag when cloning repository

sjchmiela:
# Why

We want project directories to not only contain information of current commit, but also branch or a tag. We need this for `eas-cli update --auto`.

# How

Previously, we were fetching and checking out a `gitRef` which theoretically could be a commit hash, branch name or a tag, but from looking at `www` it is always a commit hash (which I think is good).

From now on, a Git project archive config will contain:
- `gitCommitHash` — THE commit to fetch and checkout
- `gitRef` — an optional spec like `refs/heads/main`, `refs/tags/v1.0` (or `main` even). We will parse this ref and, when fetching, set a destination for the commit via a [refspec](https://www.atlassian.com/git/tutorials/refs-and-the-reflog#refspecs), similarly to [how GitHub does it](https://github.com/actions/checkout/blob/44c2b7a8a4ea60a981eaca3cf939b5f4305c123b/src/ref-helper.ts#L13).

So:
- if `gitRef` is empty, we end up in a detached HEAD state
- if `gitRef` is a (e.g.) `main` branch (`refs/heads/main`):
	- we fetch `commit` telling Git it's expected to be on a remote `main` branch
	- we checkout a new local `main` branch off the remote `main` branch (which points to the `commit`)
- if `gitRef` is a (e.g.) `v1.0` tag:
	- we fetch `commit` telling Git it's a `v1.0` tag
	- we checkout the `v1.0` tag
- if `gitRef` is something we can't parse we assume it's a branch name (`staging`)
	- we fetch `commit`
	- we checkout a new `staging` branch on the commit

Once this is merged and live, I'll submit a pull request to `www` that will add the `gitCommitHash` alongside `gitRef`.

# Test Plan

I have built `eas-build` with these changes, compiled `worker` and triggered 3 manual and 3 automatic GitHub builds, one for each: commit, branch, tag. All resolved ref correctly.

Zrzut ekranu 2024-05-3 o 23 30 50

### DIFF
--- a/packages/build-tools/src/common/projectSources.ts
+++ b/packages/build-tools/src/common/projectSources.ts
@@ -2,7 +2,8 @@ import path from 'path';
 
 import spawn from '@expo/turtle-spawn';
 import fs from 'fs-extra';
-import { ArchiveSourceType, Job } from '@expo/eas-build-job';
+import nullthrows from 'nullthrows';
+import { ArchiveSource, ArchiveSourceType, Job } from '@expo/eas-build-job';
 import { bunyan } from '@expo/logger';
 import downloadFile from '@expo/downloader';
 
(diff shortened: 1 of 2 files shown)


sjchmiela:
Just tested this with updated `www`.

For a build from a pull request got:
- `gitCommitHash = 1d1eb256fa6b1572fc4f37642a48fa105454301c`
- `gitRef = sjchmiela-patch-2`

The `build` directory was at the right commit on `sjchmiela-patch-2` branch. `git show-ref` showed remote tags and and all remote branches, not sure what's up with that, most probably since `gitRef` wasn't in qualified form, `git fetch` decided to fetch all remote refs.

For a job run from a push got:
- `gitCommitHash = 1d1eb256fa6b1572fc4f37642a48fa105454301c`
- `gitRef = refs/heads/sjchmiela-patch-2`

The `build` directory was at the right commit on `sjchmiela-patch-2` branch with no additional `show-ref` entries.

FiberJW (approved):
This makes sense to me! I'm leaving the final technical review to Szymon

sjchmiela:
Actually after thinking about this a little more I think using refspecs is unnecessarily complicated and we can just checkout the commit hash and then create a branch or a tag based on the Git ref.

When submitting this as is my hope was:
- we get an upstream set for free
- the logic is similar to that of GitHub's checkout so there's less unexpected differences

But I think it makes it harder to reason about. Will simplify tomorrow!

packages/build-tools/src/common/projectSources.ts
```diff
@@ -23,34 +24,74 @@ export async function prepareProjectSourcesAsync<TJob extends Job>(
...
+          await spawn('git', ['tag', name], { cwd: destinationDirectory });
+          break;
+        }
+        // gitRef for a branch may come as:
+        // - qualified ref (e.g. refs/heads/feature/add-icon), detected as "branch" for a push,
+        // - unqualified ref (e.g. feature/add-icon), detected as "other" for a pull request.
+        case 'branch':
+        case 'other': {
+          await spawn('git', ['checkout', '-b', name], { cwd: destinationDirectory });
+          break;
```
szdziedzic, line 83: Is it possible to pull/fetch a branch in a way so we don't need to create it locally 🤔

sjchmiela: Hm, so:
- we could do what GitHub is doing — pull a commit pointing to a remote branch, then create a new local branch pointing to the remote branch head — this feels overly complicated though
- we could do something slightly simpler — pull a commit pointing to a local branch (thus creating it via a Git ref spec like `git fetch +${commit}:${branch}`. This would create the branch locally behind the scenes. (I believe this solution is what you're suggesting?)

Do you think it would be better? At first I wasn't a fan of introducing the Git ref spec concept to here `+${commit}:${ref}`, but now after playing with it a bit more I can see the benefit of having only two commands:

- `git fetch +${commit}:${ref}`
- `git checkout ${ref || commit}`

I have just tested both commands work for all: qualified branch (`refs/heads/main`) and tag (`refs/tags/v1.0`) refs and an unqualified branch ref (`main`).

szdziedzic: both approaches seem good to me, I just wanted to understand why and how better

szdziedzic (approved):

PR #402, opened 2024-05-20
[eas-build-job][build-tools] move repack overrides to Turtle and add `BuildMode.REPACK`

szdziedzic:
# Why

Follow up to discussion in https://github.com/expo/universe/pull/15382

Move repack overrides to Turtle and add `BuildMode.REPACK`

# How

Move repack overrides to Turtle and add `BuildMode.REPACK`

# Test Plan

tests

### DIFF
--- a/packages/build-tools/src/context.ts
+++ b/packages/build-tools/src/context.ts
@@ -98,9 +98,11 @@ export class BuildContext<TJob extends Job = Job> {
     this.cacheManager = options.cacheManager;
     this._uploadArtifact = options.uploadArtifact;
     this.reportError = options.reportError;
+
+    const shouldApplyRepackOverrides = job.platform && job.mode === BuildMode.REPACK;
     this._job = {
       ...job,
-      ...(job.platform && job.mode === BuildMode.REPACK
+      ...(shouldApplyRepackOverrides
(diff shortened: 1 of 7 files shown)


packages/build-tools/src/context.ts
```diff
@@ -98,7 +98,18 @@ export class BuildContext<TJob extends Job = Job> {
     this.cacheManager = options.cacheManager;
     this._uploadArtifact = options.uploadArtifact;
     this.reportError = options.reportError;
-    this._job = job;
+
+    const shouldApplyRepackOverrides = job.platform && job.mode === BuildMode.REPACK;
```
sjchmiela, line 102: Why do we need the `job.platform` check?

sjchmiela (approved):

packages/build-tools/src/context.ts
szdziedzic: Because `Generic.Job` doesn't support `mode`

sjchmiela: Ah, sorry, I thought I'm looking at `BuildContext` in `launcher` where `job` is guaranteed to be `BuildJob`.

PR #405, opened 2024-06-05
[build-tools] resolve default values for `eas/calculate_eas_update_runtime_version` function just when executing it

szdziedzic:
# Why

https://exponent-internal.slack.com/archives/C04PTQN8UTY/p1717496873513409

Default values for build step inputs are provided at the beginning of the build process. At this point in time not all of the `job` data is updated for GH triggered build. It happens in https://github.com/expo/eas-build/blob/d70f544b51913e923803724b9581d830d641186e/packages/build-tools/src/steps/functions/resolveBuildConfig.ts#L33-L42 step. That causes the default value for `workflow` to be invalid.

# How

Make inputs non-required but if the input is not provided fall back to `job` values when executing the step (then `job` values should be correct at this point in time)

Add validation

# Test Plan

Tests

### DIFF
--- a/packages/build-tools/src/steps/easFunctions.ts
+++ b/packages/build-tools/src/steps/easFunctions.ts
@@ -51,7 +51,7 @@ export function getEasFunctions(ctx: CustomBuildContext): BuildFunction[] {
     createInstallPodsBuildFunction(),
     createSendSlackMessageFunction(),
 
-    calculateEASUpdateRuntimeVersionFunction(ctx),
+    calculateEASUpdateRuntimeVersionFunction(),
 
     createRepackBuildFunction(),
   ];
--- a/packages/build-tools/src/steps/functionGroups/build.ts
+++ b/packages/build-tools/src/steps/functionGroups/build.ts
@@ -68,11 +68,10 @@ function createStepsForIosSimulatorBuild({
   globalCtx,
   buildToolsContext,
 }: HelperFunctionsInput): BuildStep[] {
-  const calculateEASUpdateRuntimeVersion = calculateEASUpdateRuntimeVersionFunction(
-    buildToolsContext
-  ).createBuildStepFromFunctionCall(globalCtx, {
-    id: 'calculate_eas_update_runtime_version',
-  });
+  const calculateEASUpdateRuntimeVersion =
+    calculateEASUpdateRuntimeVersionFunction().createBuildStepFromFunctionCall(globalCtx, {
+      id: 'calculate_eas_update_runtime_version',
+    });
   const installPods = createInstallPodsBuildFunction().createBuildStepFromFunctionCall(globalCtx, {
     workingDirectory: './ios',
   });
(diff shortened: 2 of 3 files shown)


Kudo (approved):
looks good from calculateEASUpdateRuntimeVersionFunction but i have a sense that we might better to solve GH workflow difference with more systematic way. in case other build functions have the similar problems in the future.

szdziedzic:
>looks good from calculateEASUpdateRuntimeVersionFunction but i have a sense that we might better to solve GH workflow difference with more systematic way. in case other build functions have the similar problems in the future.

Yeah, I agree

The decision to [handle GH builds this way](https://www.notion.so/expo/Building-from-GitHub-proposal-c76243fbcde24a3b83af3849e690197f?pvs=4) keeps biting us 😬

wschurman:
I wonder if this is susceptible to the same issue that these are trying to solve:
https://github.com/expo/expo/pull/28403
https://github.com/expo/eas-build/pull/384 (mainly this change)

The issue was that the workflow detection wasn't accurate since the EAS build upload didn't contain the same folders and workflow detection happened after prebuild (thus, turning a managed project into a generic project).

wschurman:
When does the `resolve_build_config` run? Is there a way to guarantee it runs before prebuild? How does `resolve_build_config` resolve workflow?

wschurman (approved):
See previous comments for some questions. There's a chance that fingerprint may be just fundamentally incompatible with our current github implementation if any of the questions are the case.

Either way, this PR looks good.

szdziedzic:
>The issue was that the workflow detection wasn't accurate since the EAS build upload didn't contain the same folders and workflow detection happened after prebuild (thus, turning a managed project into a generic project).

>When does the resolve_build_config run? Is there a way to guarantee it runs before prebuild? How does resolve_build_config resolve workflow?

In the case of the bug reported by the user, they were using the `eas/build` function group. We fully control the contents and order of execution within this function group therefore we can guarantee a certain order of operations:
- eas/checkout
- eas/use_npm_token
- eas/install_node_modules
- eas/resolve_build_config
- **eas/calculate_eas_update_runtime_version**
- **eas/prebuild**
- ...
Within this function group, the `eas/calculate_eas_update_runtime_version` function will always be executed before prebuild.

...But users can also freely use provided custom functions on their own, so technically if somebody uses `eas/calculate_eas_update_runtime_version` after the prebuild they can mess up their build flow. We should make documentation of this function clear about how it works and when one should use it to get the correct results. We can also detect if `eas/prebuild` is used before `eas/calculate_eas_update_runtime_version` and print a warning then.

ludwig-pro:
@szdziedzic I don't know if it's a side effect but I've tried your fix and I'm having a problem with the step : Find and upload build artifacts => `There are no files matching pattern "ios/build/*.ipa"`

 https://expo.dev/accounts/stadion/projects/newcastle-app/builds/562602ba-29d0-45d5-8116-a3d7b4e0ba71

PR #407, opened 2024-06-11
[build-tools] [ENG-12652] display fingerprint sources diff is not the same

kadikraman:
### Related PRs
- https://github.com/expo/universe/pull/15517
- https://github.com/expo/eas-cli/pull/2422
- 👉 https://github.com/expo/eas-build/pull/407
- https://github.com/expo/eas-build/pull/421
- https://github.com/expo/turtle-v2/pull/1893
- https://github.com/expo/universe/pull/15592
-  https://github.com/expo/terraform-iam/pull/303

# Why

It's very difficult to debug why the local and remote fingerprint are different.

Currently the only way to do it would be manual: add step to print out the remote sources, calculate local fingerprint, diff the two. Ideally we'd like to show the diff in the build step instead.

# How

Save the local fingerprint in a google cloud bucket and show a diff if the fingerprints are different.

# Test Plan

Ran turtle and www locally with changes from the these packages manually linked to turtle.

Screenshot 2024-06-20 at 15 05 04

### DIFF
--- a/packages/build-tools/package.json
+++ b/packages/build-tools/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@expo/build-tools",
-  "version": "1.0.117",
+  "version": "1.0.119",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "files": [
@@ -24,14 +24,14 @@
   "dependencies": {
     "@expo/config": "8.5.4",
     "@expo/config-plugins": "7.8.4",
-    "@expo/downloader": "1.0.117",
-    "@expo/eas-build-job": "1.0.117",
+    "@expo/downloader": "1.0.118",
+    "@expo/eas-build-job": "1.0.119",
     "@expo/env": "^0.3.0",
     "@expo/logger": "1.0.117",
     "@expo/package-manager": "1.5.2",
     "@expo/plist": "^0.1.3",
     "@expo/repack-app": "0.0.6",
-    "@expo/steps": "1.0.117",
+    "@expo/steps": "1.0.119",
     "@expo/template-file": "1.0.117",
     "@expo/turtle-spawn": "1.0.117",
     "@expo/xcpretty": "^4.3.1",
--- a/packages/build-tools/package.json
+++ b/packages/build-tools/package.json
@@ -27,7 +27,6 @@
     "@expo/downloader": "1.0.117",
     "@expo/eas-build-job": "1.0.117",
     "@expo/env": "^0.3.0",
-    "@expo/fingerprint": "^0.9.0",
     "@expo/logger": "1.0.117",
     "@expo/package-manager": "1.5.2",
     "@expo/plist": "^0.1.3",
--- a/packages/build-tools/package.json
+++ b/packages/build-tools/package.json
@@ -27,6 +27,7 @@
     "@expo/downloader": "1.0.117",
     "@expo/eas-build-job": "1.0.117",
     "@expo/env": "^0.3.0",
+    "@expo/fingerprint": "^0.9.0",
     "@expo/logger": "1.0.117",
     "@expo/package-manager": "1.5.2",
     "@expo/plist": "^0.1.3",
--- a/packages/build-tools/src/builders/ios.ts
+++ b/packages/build-tools/src/builders/ios.ts
@@ -107,7 +107,6 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
     await ctx.runBuildPhase(BuildPhase.CONFIGURE_EXPO_UPDATES, async () => {
       await configureExpoUpdatesIfInstalledAsync(ctx, {
         resolvedRuntimeVersion: resolvedExpoUpdatesRuntimeVersion?.runtimeVersion ?? null,
-        resolvedFingerprintSources: resolvedExpoUpdatesRuntimeVersion?.fingerprintSources ?? null,
       });
     });
 
--- a/packages/build-tools/src/builders/ios.ts
+++ b/packages/build-tools/src/builders/ios.ts
@@ -106,7 +106,8 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
 
     await ctx.runBuildPhase(BuildPhase.CONFIGURE_EXPO_UPDATES, async () => {
       await configureExpoUpdatesIfInstalledAsync(ctx, {
-        resolvedRuntimeVersion: resolvedExpoUpdatesRuntimeVersion,
+        resolvedRuntimeVersion: resolvedExpoUpdatesRuntimeVersion?.runtimeVersion ?? null,
+        resolvedFingerprintSources: resolvedExpoUpdatesRuntimeVersion?.fingerprintSources ?? null,
       });
     });
 
@@ -118,8 +119,13 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
         scheme,
         buildConfiguration,
         entitlements,
-        ...(resolvedExpoUpdatesRuntimeVersion
-          ? { extraEnv: { EXPO_UPDATES_FINGERPRINT_OVERRIDE: resolvedExpoUpdatesRuntimeVersion } }
+        ...(resolvedExpoUpdatesRuntimeVersion?.runtimeVersion
+          ? {
+              extraEnv: {
+                EXPO_UPDATES_FINGERPRINT_OVERRIDE:
+                  resolvedExpoUpdatesRuntimeVersion?.runtimeVersion,
+              },
+            }
           : null),
       });
     });
--- a/packages/build-tools/src/builders/ios.ts
+++ b/packages/build-tools/src/builders/ios.ts
@@ -107,6 +107,7 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
     await ctx.runBuildPhase(BuildPhase.CONFIGURE_EXPO_UPDATES, async () => {
       await configureExpoUpdatesIfInstalledAsync(ctx, {
         resolvedRuntimeVersion: resolvedExpoUpdatesRuntimeVersion?.runtimeVersion ?? null,
+        resolvedFingerprintSources: resolvedExpoUpdatesRuntimeVersion?.fingerprintSources ?? null,
       });
     });
 
--- a/packages/build-tools/src/steps/functions/calculateEASUpdateRuntimeVersion.ts
+++ b/packages/build-tools/src/steps/functions/calculateEASUpdateRuntimeVersion.ts
@@ -76,7 +76,9 @@ export function calculateEASUpdateRuntimeVersionFunction(): BuildFunction {
         env,
       });
       if (resolvedRuntimeVersion) {
-        outputs.resolved_eas_update_runtime_version.set(resolvedRuntimeVersion);
+        outputs.resolved_eas_update_runtime_version.set(
+          resolvedRuntimeVersion.runtimeVersion ?? undefined
+        );
       } else {
         stepCtx.logger.info('Skipped because EAS Update is not configured');
       }
--- a/packages/build-tools/src/utils/expoUpdates.ts
+++ b/packages/build-tools/src/utils/expoUpdates.ts
@@ -88,7 +88,10 @@ export async function configureExpoUpdatesIfInstalledAsync(
   {
     resolvedRuntimeVersion,
     resolvedFingerprintSources,
-  }: { resolvedRuntimeVersion: string | null; resolvedFingerprintSources?: object[] | null }
+  }: {
+    resolvedRuntimeVersion: string | null;
+    resolvedFingerprintSources?: FingerprintSource[] | null;
+  }
 ): Promise<void> {
   const expoUpdatesPackageVersion = await getExpoUpdatesPackageVersionIfInstalledAsync(
     ctx.getReactNativeProjectDirectory(),
@@ -108,7 +111,7 @@ export async function configureExpoUpdatesIfInstalledAsync(
 
     if (ctx.metadata?.fingerprintSource && resolvedFingerprintSources && resolvedRuntimeVersion) {
       try {
-        const fingerprintSource = ctx.metadata?.fingerprintSource;
+        const fingerprintSource = ctx.metadata.fingerprintSource;
 
         let localFingerprint: Fingerprint | null = null;
 
@@ -124,7 +127,7 @@ export async function configureExpoUpdatesIfInstalledAsync(
         if (localFingerprint) {
           const easFingerprint = {
             hash: resolvedRuntimeVersion,
-            sources: resolvedFingerprintSources as FingerprintSource[],
+            sources: resolvedFingerprintSources,
           };
           const changes = diffFingerprints(localFingerprint, easFingerprint);
           if (changes.length) {
@@ -189,7 +192,10 @@ export async function resolveRuntimeVersionForExpoUpdatesIfConfiguredAsync({
   workflow: Workflow;
   logger: bunyan;
   env: BuildStepEnv;
-}): Promise<{ runtimeVersion: string | null; fingerprintSources: object[] | null } | null> {
+}): Promise<{
+  runtimeVersion: string | null;
+  fingerprintSources: FingerprintSource[] | null;
+} | null> {
   const expoUpdatesPackageVersion = await getExpoUpdatesPackageVersionIfInstalledAsync(cwd, logger);
   if (expoUpdatesPackageVersion === null) {
     return null;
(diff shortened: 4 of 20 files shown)


packages/eas-build-job/src/metadata.ts
```diff
@@ -54,6 +54,11 @@ export type Metadata = {
    */
   runtimeVersion?: string;
 
+  /**
+   * Sources used to calculate the project fingerprint (for Expo Updates)
```
wschurman, line 58:
```suggestion
   * Sources used to calculate the project fingerprint and their hashes.
```

packages/build-tools/package.json
```diff
@@ -27,6 +27,7 @@
     "@expo/downloader": "1.0.117",
     "@expo/eas-build-job": "1.0.117",
     "@expo/env": "^0.3.0",
+    "@expo/fingerprint": "^0.9.0",
```
wschurman, line 30: I've been hesitant to hardcode this package dependency since we could see version incompatibility issues if the version used by the expo-updates in the project differs from this and changes have been made to one or the other. Put another way, since the fingerprint depends on the version of the `@expo/fingerprint` package, we generally need to ensure that any fingerprint calculations we do from the build service are done via RPC calls into the app's expo-update CLI. 

Since we're just using the `diffFingerprints` it probably doesn't matter quite as much (that function shouldn't ever need change), but the risk is that somewhere down the line we accidentally start using other functions from the package here in the build service and forget that it needs to be versioned through the expo-updates CLI.

So, my suggestion is that we don't add this dependency and instead just copy the `diffFingerprints` function into `build-tools`.

wschurman (changes requested):
The changes and the addition of the log looks great. Just one note about using the package.

packages/build-tools/package.json
kadikraman: Good idea! I wasn't a fan of adding fingerprint as a dependency either, but couldn't think of an alternative. I was intended to use the cli initially, but all the [cli options](https://github.com/expo/expo/tree/main/packages/%40expo/fingerprint#cli-usage) rely on passing in the _project path_ first so it doesn't support diffing two fingerprints. Copying the function over seems like a good compromise 👍

packages/build-tools/src/utils/expoUpdates.ts
```diff
@@ -96,6 +100,38 @@ export async function configureExpoUpdatesIfInstalledAsync(
...
+                return value;
+              }
+            }
+            return value;
+          },
+          ' '
+        )
+      );
+    } else {
+      ctx.logger.warn(`Could not find fingerprint sources`);
```
Kudo, line 133: we should maybe consider if runtimeVersion.policy!=fingerprint case here. in this case i guess the `ctx.metadata.runtimeVersion` would be [`file:fingerprint`](https://github.com/expo/expo/blob/88b74f7ff00d37b2e44b1fa6e6dc764cc028b1d4/packages/%40expo/config-plugins/src/utils/Updates.ts#L16)

Kudo (changes requested):

packages/build-tools/src/steps/functions/calculateEASUpdateRuntimeVersion.ts
```diff
@@ -76,7 +76,9 @@ export function calculateEASUpdateRuntimeVersionFunction(): BuildFunction {
         env,
       });
       if (resolvedRuntimeVersion) {
-        outputs.resolved_eas_update_runtime_version.set(resolvedRuntimeVersion);
+        outputs.resolved_eas_update_runtime_version.set(
+          resolvedRuntimeVersion.runtimeVersion ?? undefined
+        );
```
szdziedzic, line 81: This function will probably need to have a second output `resolved_eas_update_fingerprint_sources` so we do exactly the same thing for custom builds as for standard builds

szdziedzic (changes requested):

packages/build-tools/src/steps/functions/calculateEASUpdateRuntimeVersion.ts
kadikraman: The fingerprints are now stored in a GC bucket.
I think this means that the fingerprint diffing won't work on a local build because you won't be able to access the signed url (due to cors?).
I imagine it's ok to have this as an EAS cloud only feature though.

packages/build-tools/src/utils/expoUpdates.ts
kadikraman: Changed the logic here, we only compare fingerprint sources if both local and eas have them, and no warning is printed if they don't exist.

packages/build-tools/package.json
kadikraman: Copied Kudo's `diffFingerprints` function https://github.com/expo/expo/pull/29709

packages/eas-build-job/src/metadata.ts
kadikraman: That field is now removed

packages/build-tools/src/utils/fingerprint.ts
```diff
@@ -0,0 +1,175 @@
+export type FingerprintSource = HashSource & {
+  /**
+   * Hash value of the `source`.
+   * If the source is excluding by `Options.dirExcludes`, the value will be null.
```
ide, line 4: Would be useful to document what hash algorithm is used.

packages/build-tools/src/utils/expoUpdates.ts
```diff
@@ -88,14 +93,38 @@ export async function configureExpoUpdatesIfInstalledAsync(
...
+    ctx.logger.warn(`Runtime version on your local machine: ${ctx.metadata.runtimeVersion}`);
+    ctx.logger.warn(`Runtime version calculated on EAS: ${appConfigRuntimeVersion}`);
+
+    if (
+      ctx.metadata?.projectFingerprintUrl &&
+      resolvedFingerprintSources &&
+      resolvedRuntimeVersion
+    ) {
+      try {
+        const result = await fetch(ctx.metadata?.projectFingerprintUrl);
```
wschurman, line 109: Ah, I see, this is where the URL is used from the build job.

@sjchmiela and @szdziedzic - What's the best way to access a file from within a GCS bucket in build tools is we only have the bucket key? Is it okay to put temporary data in the build job object like this expiring URL? Or is there a way to call www from build-tools to generate a fresh signed URL from a key?

Maybe this is kinda like the `secrets.robotAccessToken` which is also temporary data within the build job?

@kadikraman - if we do decide to put the temporary URL into the build job we'll need to increase how long the URL is good for to longer than the max queue time plus max build time to ensure the URL can still be fetched even in the worst case.

wschurman (changes requested):
Back to your queue until we figure out what to do about the expiring URL.

packages/build-tools/src/steps/functions/calculateEASUpdateRuntimeVersion.ts
szdziedzic: Yeah, that's true, in its current form, it wouldn't work for local builds because there would be no place to generate the signed URL for the fingerprint to be downloaded (local builds don't go through Turtle and WWW). 

Local builds use something called local-build-plugin
https://github.com/expo/eas-build/blob/69bbd3e7490b75a422e7439567d9627914d851a4/packages/local-build-plugin/src/main.ts#L1-L23

You can add support for it by basically not uploading the fingerprint to GCS when running the local build but copying it to some temp dir locally and then getting it from there during the build process.

That's what we do for project archive sources.

This will require supporting `PATH` source for fingerprint.

packages/eas-build-job/src/metadata.ts
```diff
@@ -173,6 +183,8 @@ export const MetadataSchema = Joi.object({
   credentialsSource: Joi.string().valid('local', 'remote'),
   sdkVersion: Joi.string(),
   runtimeVersion: Joi.string(),
+  fingerprintGCSBucketKey: Joi.string(),
+  projectFingerprintUrl: Joi.string(),
```
szdziedzic, line 187: You would need to add 
```ts
export const FingerprintSourceSchema = Joi.object<FingerprintSource>({
  type: Joi.string()
    .valid(...Object.values(FingerprintSourceType))
    .required(),
})
  .when(Joi.object({ type: FingerprintSourceType.GCS }).unknown(), {
    then: Joi.object({
      type: Joi.string().valid(ArchiveSourceType.GCS).required(),
      bucketKey: Joi.string().required(),
    }),
  })
  .when(Joi.object({ type: ArchiveSourceType.PATH }).unknown(), {
    then: Joi.object({
      type: Joi.string().valid(ArchiveSourceType.PATH).required(),
      path: Joi.string().required(),
    }),
  });
```

And then do 
```suggestion
  fingerprintSource: FingerprintSourceSchema,
```

packages/build-tools/src/utils/expoUpdates.ts
szdziedzic, line 109: We will need to handle both `FingerprintSourceType.URL` and `FingerprintSourceType.PATH` here

szdziedzic (changes requested):
One more PR in the `turtle-v2` repo will be needed to add support for generating the pre-signed download URL for the fingerprint source https://github.com/expo/eas-build/pull/407#discussion_r1650611496

Generally, we will need to support 3 kinds of fingerprint sources:
1. `GCS` - type created by EAS CLI when starting cloud build, it contains data about `bucketKey` and is passed to WWW and then to Turtle launcher which will generate the pre-signed URL based on it
2. `URL` - pre-signed URL generated by launcher, passed to cloud worker
3. `PATH` - for local builds

Let me know if it's not clear or if you have any more questions

packages/build-tools/src/utils/expoUpdates.ts
kadikraman: This is now done in the turtle PR.

packages/build-tools/src/utils/fingerprint.ts
kadikraman: I'll add a comment to this file - the function is a direct copy/paste from the fingerprint repo where it also has tests.
https://github.com/expo/expo/pull/29709

packages/build-tools/src/utils/expoUpdates.ts
kadikraman: Could I handle the `FingerprintSourceType.PATH` in a separate PR?

packages/build-tools/src/utils/expoUpdates.ts
```diff
@@ -88,14 +94,36 @@ export async function configureExpoUpdatesIfInstalledAsync(
...
-    ctx.logger.warn(
-      "If you're using conditional app configs, e.g. depending on an environment variable, make sure to set the variable in eas.json or configure it with EAS Secret."
-    );
+    ctx.logger.warn('Runtime version mismatch');
+    ctx.logger.warn(`Runtime version on your local machine: ${ctx.metadata.runtimeVersion}`);
+    ctx.logger.warn(`Runtime version calculated on EAS: ${appConfigRuntimeVersion}`);
+
+    if (ctx.metadata?.fingerprintSource && resolvedFingerprintSources && resolvedRuntimeVersion) {
+      try {
+        const fingerprintSource = ctx.metadata?.fingerprintSource;
```
wschurman, line 106:
```suggestion
        const fingerprintSource = ctx.metadata.fingerprintSource;
```

packages/build-tools/src/utils/expoUpdates.ts
```diff
@@ -88,14 +94,36 @@ export async function configureExpoUpdatesIfInstalledAsync(
...
+
+        if (fingerprintSource.type === FingerprintSourceType.URL) {
+          const result = await fetch(fingerprintSource.url);
+          const localFingerprint = await result.json();
+          const easFingerprint = {
+            hash: resolvedRuntimeVersion,
+            sources: resolvedFingerprintSources as FingerprintSource[],
```
wschurman, line 113: Might be cleaner to hoist this case all the way up to `resolveRuntimeVersionAsync` and use `FingerprintSource[]` instead of `object[]` as types everywhere. 

Optionally could also add validation in `resolveRuntimeVersionAsync` to check that `runtimeVersionResult.fingerprintSources` adheres to the expected type before returning it using a library like `zod`. This would be a good stacked PR though as it's a bit outside the scope of this PR.

wschurman (approved):
A few inline comments.

packages/build-tools/src/builders/ios.ts
```diff
@@ -106,7 +106,7 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
 
     await ctx.runBuildPhase(BuildPhase.CONFIGURE_EXPO_UPDATES, async () => {
       await configureExpoUpdatesIfInstalledAsync(ctx, {
-        resolvedRuntimeVersion: resolvedExpoUpdatesRuntimeVersion,
+        resolvedRuntimeVersion: resolvedExpoUpdatesRuntimeVersion?.runtimeVersion ?? null,
```
szdziedzic, line 109:
```suggestion
        resolvedRuntimeVersion: resolvedExpoUpdatesRuntimeVersion?.runtimeVersion ?? null,
        resolvedFingerprintSources: resolvedExpoUpdatesRuntimeVersion?.fingerprintSources ?? null,
```

Shouldn't this also have  `resolvedFingerprintSources` passed?

packages/build-tools/src/utils/expoUpdates.ts
```diff
@@ -88,14 +94,36 @@ export async function configureExpoUpdatesIfInstalledAsync(
...
+          };
+          const changes = diffFingerprints(localFingerprint, easFingerprint);
+          if (changes.length) {
+            ctx.logger.warn('Difference between local and EAS fingerprints:');
+            ctx.logger.warn(stringifyFingerprintDiff(changes));
+          }
+        } else if (fingerprintSource.type === FingerprintSourceType.PATH) {
+          // TODO(Kadi): handle local fingerprint path files
```
szdziedzic, line 121: I think that the only thing to do here when it comes to loading files will be
```suggestion
          const localFingerprint = await fs.readJson(fingerprintSource.path);
          // ... additional parsing and diffing as above
```

packages/build-tools/src/utils/expoUpdates.ts
```diff
@@ -88,14 +94,36 @@ export async function configureExpoUpdatesIfInstalledAsync(
...
+        }
```
szdziedzic, line 122:
```suggestion
        } else { 
          ctx.logger.warn(`Invalid fingerprint source type: ${fingerprintSource.type}`);
        }
```

packages/build-tools/src/utils/expoUpdates.ts
```diff
@@ -5,6 +5,8 @@ import semver from 'semver';
 import { ExpoConfig } from '@expo/config';
 import { bunyan } from '@expo/logger';
 import { BuildStepEnv } from '@expo/steps';
+import fetch from 'node-fetch';
+import { FingerprintSourceType } from '@expo/eas-build-job/dist/metadata';
```
szdziedzic, line 9:
```suggestion
import { FingerprintSourceType } from '@expo/eas-build-job';
```

packages/build-tools/src/utils/fingerprint.ts
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   * Sources and their hash values to generate a fingerprint
+   */
+  sources: FingerprintSource[];
+
+  /**
+   * The final hash value of the whole fingerprint
+   */
+  hash: string;
+}
```
szdziedzic, line 29: Can we import types like this from `@expo/fingerprint`? I'm asking because it seems to me that it might be easier to keep types/functions in sync just by bumping the package version (assuming the package will be backward compatible).

szdziedzic (changes requested):

packages/build-tools/src/utils/fingerprint.ts
kadikraman: Discussed importing this function with @Kudo and the summary of the discussion is that we can't because we don't want to add `@expo/fingerprint` as a dependency.

packages/build-tools/src/builders/ios.ts
kadikraman: Yes! Good spot. You can tell I tend to use the Android build for local testing 😳

szdziedzic (approved):
thanks!

packages/build-tools/src/utils/expoUpdates.ts
```diff
@@ -88,14 +102,43 @@ export async function configureExpoUpdatesIfInstalledAsync(
...
+          const easFingerprint = {
+            hash: resolvedRuntimeVersion,
+            sources: resolvedFingerprintSources,
+          };
+          const changes = diffFingerprints(localFingerprint, easFingerprint);
+          if (changes.length) {
+            ctx.logger.warn('Difference between local and EAS fingerprints:');
+            ctx.logger.warn(stringifyFingerprintDiff(changes));
+          }
+        }
```
Kudo, line 137: nit: maybe move this code to a separate function

Kudo (approved):
overall good, just a nit comment to move code out to separate function to keep the original code cleaner.

PR #408, opened 2024-06-13
[steps] [ENG-11679] Fix escape characters in step inputs

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-11679/investigate-incorrect-escaping-of-special-characters-in-custom-builds
When passing a string input into a step the escape characters used to get doubly escaped. JS can work around it and treat them as normal, but when used with external services, like Slack API, it causes problems. For example, Slack messages sent via the `send_slack_message` step had `\\n` in them instead of the new-line character.

# How

The input value is JSON.stringified for serialization and therefore needs to be JSON.parsed to remove the doubly escaped characters. This fixes all escape characters, not only `\n` as before

# Test Plan
## Automated tests
Automated tests pass

## Manual tests
| Before | After |
| - | - |
| Screenshot 2024-06-13 at 14 32 52 | Screenshot 2024-06-13 at 14 33 03 |

### DIFF
--- a/packages/steps/src/BuildStepInput.ts
+++ b/packages/steps/src/BuildStepInput.ts
@@ -132,22 +132,7 @@ export class BuildStepInput<
       );
       returnValue = this.parseInputValueToAllowedType(valueInterpolatedWithOutputsAndGlobalContext);
     }
-    return this.fixEscapeCharacters(returnValue);
-  }
-
-  private fixEscapeCharacters(
-    input: BuildStepInputValueTypeWithRequired<T, R>
-  ): BuildStepInputValueTypeWithRequired<T, R> {
-    if (typeof input === 'string') {
-      return input.replace(/\\n/g, '\n') as BuildStepInputValueTypeWithRequired<T, R>;
-    } else if (typeof input === 'object') {
-      for (const property of Object.keys(input)) {
-        input[property] = this.fixEscapeCharacters(input[property]);
-      }
-      return input;
-    } else {
-      return input;
-    }
+    return returnValue;
   }
 
   public get rawValue(): BuildStepInputValueType<T> | undefined {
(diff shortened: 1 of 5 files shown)


packages/steps/src/BuildStepInput.ts
```diff
@@ -227,6 +212,19 @@ export class BuildStepInput<
     }
   }
 
+  private parseInputValueToString(value: string): string {
+    let parsedValue = value;
+    try {
+      parsedValue = JSON.parse(`"${value}"`);
+    } catch (err) {
+      // in case of SyntaxError return raw string
```
szdziedzic, line 220: When can it occur?

szdziedzic (approved):

packages/steps/src/BuildStepInput.ts
radoslawkrzemien: It came up with manual testing after I fixed a failing test by also parsing the strings that get interpolated into the object properties, which fixed the test.
After your question I did some more testing and it turns out that it was the test that was broken and that additional parsing is not needed since the injected text is already parsed but the object property in which it gets injected is not and when the property itself is parsed later it causes the syntaxError.
I have removed the faulty test as well as additional parsing of the string injected into the object property since the property gets parsed later as a whole. Manual testing confirms this works

packages/steps/src/BuildStepInput.ts
```diff
@@ -227,6 +212,10 @@ export class BuildStepInput<
     }
   }
 
+  private parseInputValueToString(value: string): string {
+    return JSON.parse(`"${value}"`);
+  }
```
szdziedzic, line 217: What will happen if one passes string `test\ntest2` as input? I believe it will fail, right?

Can we guarantee that we will always use `\\n` instead of `\n` in inputs 🤔?

radoslawkrzemien: I'm pretty sure it's guaranteed and all inputs are JSON.stringified.
If we want a fallback, I can bring back the SyntaxError check, in which case it will return the raw string

PR #420, opened 2024-06-24
[build-tools] fix issues with `./gradlew` being executed with inncorrect shell

szdziedzic:
# Why

Using always `bash` or `sh` seems undesired and it looks like it caused some issues for one of our users https://exponent-internal.slack.com/archives/C02123T524U/p1718651239155819

# How

Use `gradlew` as just `./gradlew` instead of `bash -c sh gradlew`.

The `bash -c sh gradlew` hack was introduced to fix permission issues when executing the file. To preserve the fix I added the `chmod +x` before.

# Test Plan

tests


szdziedzic:
@sjchmiela what do you think? This is the version from before the `fix permissions` commit

sjchmiela (approved):

PR #426, opened 2024-07-20
[create-eas-build-function] Add missing ncc dependency

douglowder:
# Why

Noticed that when creating a new build function with `create-eas-build-function`, the "build" script in the new function's `package.json` will not work unless `@vercel/ncc` is added to the package.

# How

Added the missing dependency to both templates.

# Test Plan

- CI should pass
- Create a new function, go into the function directory, and call "yarn" and "yarn build". After this change, those steps should work.

### DIFF
--- a/packages/create-eas-build-function/templates/javascript/package.json
+++ b/packages/create-eas-build-function/templates/javascript/package.json
@@ -4,6 +4,7 @@
   "main": "./build/index.js",
   "type": "commonjs",
   "devDependencies": {
+    "@vercel/ncc": "^0.38.1",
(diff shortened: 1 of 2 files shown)


sjchmiela (approved):
Ha, I remember hitting this too! As far as I remember it was explained to me it was done this way because [`ncc` installation instructions](https://github.com/vercel/ncc?tab=readme-ov-file#installation) advise to install it globally and this is what [we tell people too](https://docs.expo.dev/custom-builds/functions/#compile-the-function).

I'm good with adding this to streamline setup for developers.

PR #429, opened 2024-07-22
[1/n][ENG-12843][eas-build-job] add new steps schema

szdziedzic:
# Why

1 of n of [ENG-12843](https://linear.app/expo/issue/ENG-12843/allow-passing-whole-job-run-config-to-turtle-service-from-www-instead)

# How

Add new steps schema that are mostly similar/compatible with the [GHA schema](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions).

Allow to pass this new schema object as a part of the Generic Job object (It's allowed to pass the value, but it isn't used anywhere yet. Passing path is still required, it will change in the next PR(s)).

# Test Plan

Tests

### DIFF
--- a/packages/eas-build-job/src/customJob.ts
+++ b/packages/eas-build-job/src/customJob.ts
@@ -1,10 +1,10 @@
 import { z } from 'zod';
 
-export const CustomJobShellZ = z.enum(['sh', 'bash']);
+export const CustomJobStepShellZ = z.enum(['sh', 'bash']);
 
-export type CustomJobShell = z.infer<typeof CustomJobShellZ>;
+export type CustomJobStepShell = z.infer<typeof CustomJobStepShellZ>;
 
-export const CommonCustomJobStepZ = z.object({
+const CommonCustomJobStepZ = z.object({
   /**
    * Unique identifier for the step.
    *
@@ -48,7 +48,6 @@ export const CommonCustomJobStepZ = z.object({
 });
 
 export const CustomJobStepWithUsesFieldZ = CommonCustomJobStepZ.extend({
-  ...CommonCustomJobStepZ.shape,
   /**
    * The custom EAS function to run as a step.
    * It can be a function provided by EAS or a custom function defined by the user.
@@ -103,7 +102,7 @@ export const CustomJobStepWithRunFieldZ = CommonCustomJobStepZ.extend({
    *
    * @default 'bash'
    */
-  shell: CustomJobShellZ.optional(),
+  shell: CustomJobStepShellZ.optional(),
 });
 
 export const CustomJobStepZ = z.union([CustomJobStepWithUsesFieldZ, CustomJobStepWithRunFieldZ]);
--- a/packages/eas-build-job/src/customJob.ts
+++ b/packages/eas-build-job/src/customJob.ts
@@ -69,6 +69,9 @@ export const CustomJobFunctionStepZ = CommonCustomJobStepZ.extend({
    *  arg4: '${{ steps.step1.outputs.test }}'
    */
   with: z.record(z.unknown()).optional(),
+
+  run: z.never().optional(),
+  shell: z.never().optional(),
 }).strict();
 
 export type CustomJobFunctionStep = z.infer<typeof CustomJobFunctionStepZ>;
@@ -96,6 +99,9 @@ export const CustomJobScriptStepZ = CommonCustomJobStepZ.extend({
    * @default 'bash'
    */
   shell: z.string().optional(),
+
+  uses: z.never().optional(),
+  with: z.never().optional(),
 }).strict();
 
 export type CustomJobScriptStep = z.infer<typeof CustomJobScriptStepZ>;
--- a/packages/eas-build-job/src/customJob.ts
+++ b/packages/eas-build-job/src/customJob.ts
@@ -1,0 +1,160 @@
+import { z } from 'zod';
+
+export const CustomJobShellZ = z.enum(['sh', 'bash']);
+
+export type CustomJobShell = z.infer<typeof CustomJobShellZ>;
+
+export const CommonCustomJobStepZ = z.object({
+  /**
+   * Unique identifier for the step.
+   *
+   * @example
+   * id: 'step1'
+   */
+  id: z.string().optional(),
+  /**
+   * Expression that determines whether the step should run.
+   * Based on the GitHub Actions job step `if` field (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif).
+   *
+   * @example
+   * if: '${{ steps.step1.outputs.test == 'test' }}'
+   */
+  if: z.string().optional(),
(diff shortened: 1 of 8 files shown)


packages/eas-build-job/src/generic.ts
```diff
@@ -22,9 +23,15 @@ export namespace Generic {
   export type Job = z.infer<typeof JobZ>;
   export const JobZ = z.object({
     projectArchive: ArchiveSourceSchemaZ,
+    // TODO: remove this once we pass all of the jobs to Turtle as objects
     customBuildConfig: z.object({
       path: z.string(),
     }),
+    customJob: z
```
sjchmiela, line 30: Let's make it `steps: z.array(...).optional()` so we don't have a (custom) job inside a (generic) job?

packages/eas-build-job/src/customJob.ts
```diff
@@ -0,0 +1,135 @@
...
+   *
+   * @example
+   * name: 'Step 1'
+   */
+  name: z.string().optional(),
+  /**
+   * The working directory to run the step in.
+   *
+   * @example
+   * working-directory: './my-working-directory'
```
sjchmiela, line 30:
```suggestion
   * working_directory: './my-working-directory'
```

packages/eas-build-job/src/customJob.ts
```diff
@@ -0,0 +1,135 @@
...
+   *
+   * @default '.' (the root of the repository)
```
sjchmiela, line 32: This is not necessarily true, it depends on project settings.

packages/eas-build-job/src/customJob.ts
```diff
@@ -0,0 +1,135 @@
...
+  env: z.record(z.string()).optional(),
+});
+
+export const CustomJobFunctionStepZ = CommonCustomJobStepZ.extend({
+  /**
+   * The custom EAS function to run as a step.
+   * It can be a function provided by EAS or a custom function defined by the user.
+   *
+   * @example
+   * uses: 'eas/build'
```
sjchmiela, line 52: I think YAML reads nicer if string values are not wrapped with `'`

packages/eas-build-job/src/customJob.ts
```diff
@@ -0,0 +1,135 @@
...
+   * @example
+   * uses: 'my-custom-function'
+   */
+  uses: z.string(),
+  /**
+   * The arguments to pass to the function.
+   *
+   * @example
+   * with:
+   *  arg1: 'value1'
```
sjchmiela, line 63: Indentation is wrong on the examples (levels are nested by 1 space, here and in other places)

packages/eas-build-job/src/customJob.ts
```diff
@@ -0,0 +1,135 @@
...
+   * @example
+   * shell: 'sh'
+   *
+   * @default 'bash'
+   */
+  shell: z.string().optional(),
+
+  uses: z.never().optional(),
+  with: z.never().optional(),
+}).strict();
```
sjchmiela, line 105: Why would we need to put `.strict()` on an internal API? This might cause problems down the road as we add more stuff.

packages/eas-build-job/src/customJob.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export type CustomJobScriptStep = z.infer<typeof CustomJobScriptStepZ>;
+
+export const CustomJobStepZ = z.union([CustomJobScriptStepZ, CustomJobFunctionStepZ]);
```
sjchmiela, line 109: Can we use `discriminatedUnion` on `run` or on `uses`?

packages/eas-build-job/src/customJob.ts
```diff
@@ -0,0 +1,135 @@
+import { z } from 'zod';
+
+const CommonCustomJobStepZ = z.object({
```
sjchmiela, line 3: What do you think about stripping "custom" and "job" from the names here so we focus on these being "steps"?

packages/eas-build-job/src/generic.ts
```diff
@@ -22,9 +23,15 @@ export namespace Generic {
   export type Job = z.infer<typeof JobZ>;
   export const JobZ = z.object({
     projectArchive: ArchiveSourceSchemaZ,
+    // TODO: remove this once we pass all of the jobs to Turtle as objects
```
sjchmiela, line 26: Can you instead create a Linear task?

sjchmiela (changes requested):

packages/eas-build-job/src/customJob.ts
szdziedzic: I don't think so. I tried and it throws an error 🤔 

I found that
>z.discriminatedUnion requires the use of a z.literal or z.enum

packages/eas-build-job/src/steps.ts
```diff
@@ -0,0 +1,135 @@
...
+   */
+  with: z.record(z.unknown()).optional(),
+
+  run: z.never().optional(),
+  shell: z.never().optional(),
+});
+
+export type FunctionStep = z.infer<typeof FunctionStepZ>;
+
+export const ScriptStepZ = CommonStepZ.extend({
```
sjchmiela, line 79: Let's call it a shell step? Script is slightly ambiguous.

packages/eas-build-job/src/steps.ts
```diff
@@ -0,0 +1,135 @@
...
+ *    with:
+ *      flow_path: |
+ *        maestro/sign_in.yaml
+ *        maestro/create_post.yaml
+ *        maestro/sign_out.yaml
+ *  - run: echo Hello, world!
+ */
+export type Step = z.infer<typeof StepZ>;
+
+export const StepsZ = z.array(StepZ);
```
sjchmiela, line 133: I personally don't see huge benefit in introducing `StepsZ` — to me it hinders the fact it is an array (and writing `z.array(StepZ)`, `Step[]` in those few places where we would have used `StepsZ`, `Steps` and adds an unnecessary concept. I would inline arrays.

packages/eas-build-job/src/steps.ts
```diff
@@ -0,0 +1,135 @@
...
+   * @example
+   * id: step1
+   */
+  id: z.string().optional(),
+  /**
+   * Expression that determines whether the step should run.
+   * Based on the GitHub Actions job step `if` field (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif).
+   *
+   * @example
+   * if: ${{ steps.step1.outputs.test == 'test' }}
```
sjchmiela, line 16: I'm not sure this is how `if`s are going to look like. https://exponent-internal.slack.com/archives/C04GPRQCEBG/p1711554721112529?thread_ts=1711392256.213759&cid=C04GPRQCEBG

I would skip adding examples if we aren't sure what things are going to look like.

sjchmiela (approved):

PR #430, opened 2024-07-26
[2/3][ENG-12843][build-tools][steps][eas-build-job] add support for new steps format in `@expo/steps` library

szdziedzic:
# Why

We want to use the new steps config format to run custom jobs using the `@expo/steps` library.

# How

Add a new config parser for the new steps config format.

Allow outputs to be interpolated into build step commands

Update the `Generic.Job` schema and use a new parser when `Generic.Job.steps` are present.

# Test Plan

Add automatic tests


packages/build-tools/src/generic.ts
```diff
@@ -18,19 +17,22 @@ export async function runGenericJobAsync(ctx: BuildContext<Generic.Job>): Promis
...
+        externalFunctions: getEasFunctions(customBuildCtx),
+        externalFunctionGroups: getEasFunctionGroups(customBuildCtx),
+        steps: ctx.job.steps,
+      })
+    : new BuildConfigParser(globalContext, {
+        externalFunctions: getEasFunctions(customBuildCtx),
+        externalFunctionGroups: getEasFunctionGroups(customBuildCtx),
+        configPath: path.join(
+          customBuildCtx.projectSourceDirectory,
+          ctx.job.customBuildConfig.path
```
sjchmiela, line 33: Should this `nullthrows`?

packages/eas-build-job/src/generic.ts
```diff
@@ -39,4 +35,17 @@ export namespace Generic {
     triggeredBy: z.literal(BuildTrigger.GIT_BASED_INTEGRATION),
     loggerLevel: z.nativeEnum(LoggerLevel).optional(),
   });
+
+  export const JobZ = z.union([
+    CommonJobZ.extend({
+      customBuildConfig: z.object({
```
sjchmiela, line 41: Interesting, for some reason I have been imagining we would put `steps` _under_ `customBuildConfig`. It does make sense to keep it out though. Cool!

packages/eas-build-job/src/step.ts
```diff
@@ -65,10 +70,11 @@ export const FunctionStepZ = CommonStepZ.extend({
    *      - value1
    *   arg4: ${{ steps.step1.outputs.test }}
    */
-  with: z.record(z.unknown()).optional(),
+  with: z.record(z.union([z.string(), z.number(), z.record(z.any())])).optional(),
```
sjchmiela, line 73: What did this change help with? I'm worried limiting this here may shoot us down the road. Example — YAML coalesces booleans to `z.boolean()`, so `a: true` will fail to be recognized as valid.

Zrzut ekranu 2024-07-30 o 11 10 15

packages/eas-build-job/src/step.ts
```diff
@@ -126,3 +144,16 @@ export const StepZ = z.union([ShellStepZ, FunctionStepZ]);
  *  - run: echo Hello, world!
  */
 export type Step = z.infer<typeof StepZ>;
+
+export function validateSteps(maybeSteps: unknown): Step[] {
+  const steps = z.array(StepZ).min(1).parse(maybeSteps);
+  return steps;
+}
+
+export function isStepShellStep(step: Step): step is ShellStep {
```
sjchmiela, line 153: Hmm, is there a property that differentiates between shell and function step? It seems wrong to parse steps in the beginning and later parse them again for type differentiation.

packages/steps/src/StepsConfigParser.ts
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  protected async parseConfigToBuildStepsAndBuildFunctionByIdMappingAsync(): Promise<{
+    buildSteps: BuildStep[];
+    buildFunctionById: BuildFunctionById;
+  }> {
+    const validatedSteps = validateSteps(this.steps);
+    StepsConfigParser.validateAllFunctionsExist(validatedSteps, {
+      externalFunctionIds: this.getExternalFunctionFullIds(),
+      externalFunctionGroupIds: this.getExternalFunctionGroupFullIds(),
```
sjchmiela, line 54: I expect changing custom jobs in a way that does not require validation beforehand, but validates as it runs would be much simpler. Then all these parsers would magically go away?

sjchmiela (approved):
Great in general, two things to reconsider:
- limiting type in `with` — it breaks booleans and maybe more
- `isStepShellStep` and `isStepFunctionStep`

packages/build-tools/src/generic.ts
szdziedzic: I don't think so it's needed here 🤔 

Based on the `Generic.Job` type (and validation) we know that either `ctx.job.customBuildConfig.path` or `ctx.job.steps`

packages/eas-build-job/src/step.ts
szdziedzic: The `z.record(z.union([z.string(), z.number(), z.record(z.any())]), z.boolean())` is the current type of accepted `inputs` https://github.com/expo/eas-build/blob/ad0487b2e5609aa879c6b488360d11442a45f542/packages/steps/src/BuildStepInput.ts#L19-L27

You are right I previously missed `boolean` and it was `z.record(z.union([z.string(), z.number(), z.record(z.any())]))` instead of `z.record(z.union([z.string(), z.number(), z.record(z.any())]), z.boolean())` 👍

packages/eas-build-job/src/step.ts
szdziedzic: Yeah, there is. I will change it 🙌

PR #436, opened 2024-08-21
[eas-build] [steps] [ENG-13151] Catch errors while evaluating if step should be executed

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-13151/commented-out-lines-in-message-template-fail-the-send-slack-message
The function `step.shouldExecuteStep(hasAnyPreviousStepFailed)` was throwing an error which was not caught by the `try/catch` wrapping the actual step execution which resulted in a failed build without logs

# How

Wrapped the function determining if the step should be executed in a try/catch block. The caught error is logged, as is a message that the runner failed to evaluate if the step should be executed or not, so the user knows what's going on. The step is then skipped and the build is failed

# Test Plan

Automated tests pass

Manual testing

Before, when interpolated step didn't exist
Screenshot 2024-08-21 at 14 03 25

Now, when interpolated step don't exist
Screenshot 2024-08-21 at 14 03 55

Now, when interpolated step exists
Screenshot 2024-08-21 at 14 03 41

### DIFF
--- a/packages/steps/src/BuildWorkflow.ts
+++ b/packages/steps/src/BuildWorkflow.ts
@@ -19,7 +19,18 @@ export class BuildWorkflow {
     let maybeError: Error | null = null;
     let hasAnyPreviousStepFailed = false;
     for (const step of this.buildSteps) {
-      if (step.shouldExecuteStep(hasAnyPreviousStepFailed)) {
+      let shouldExecuteStep = false;
+      try {
+        shouldExecuteStep = step.shouldExecuteStep(hasAnyPreviousStepFailed);
+      } catch (err: any) {
+        step.ctx.logger.error({ err });
+        step.ctx.logger.error(
+          `Runner failed to evaluate if it should execute step "${step.displayName}"`
+        );
+        maybeError = maybeError ?? err;
+        hasAnyPreviousStepFailed = true;
+      }
+      if (shouldExecuteStep) {
         try {
           await step.executeAsync();
         } catch (err: any) {
--- a/packages/steps/src/BuildWorkflow.ts
+++ b/packages/steps/src/BuildWorkflow.ts
@@ -25,7 +25,7 @@ export class BuildWorkflow {
       } catch (err: any) {
         step.ctx.logger.error({ err });
         step.ctx.logger.error(
-          `Runner failed to evaluate if it should execute step "${step.displayName}"`
+          `Runner failed to evaluate if it should execute step "${step.displayName}", using step's if condition "${step.ifCondition}". This can be caused by trying to access non-existing object property. If you think this is a bug report it here: https://github.com/expo/eas-cli/issues.`
         );
         maybeError = maybeError ?? err;
         hasAnyPreviousStepFailed = true;


packages/steps/src/BuildWorkflow.ts
```diff
@@ -19,7 +19,18 @@ export class BuildWorkflow {
     let maybeError: Error | null = null;
     let hasAnyPreviousStepFailed = false;
     for (const step of this.buildSteps) {
-      if (step.shouldExecuteStep(hasAnyPreviousStepFailed)) {
+      let shouldExecuteStep = false;
+      try {
+        shouldExecuteStep = step.shouldExecuteStep(hasAnyPreviousStepFailed);
+      } catch (err: any) {
```
szdziedzic, line 25: The config that the user was using was:
```yml
- eas/send_slack_message:
        if: ${ success() }
        name: Send Slack message on success
        inputs:
          payload:
            blocks:
              - type: header
                text:
                  type: plain_text
                  text: 'Company - :apple_logo: iOS - Develop Pipeline :large_green_circle:'
                  emoji: true
              - type: section
                text:
                  type: mrkdwn
                  text: |-
                    *Build Number:* ${ eas.job.version.buildNumber }
              - type: divider
              - type: section
                text:
                  type: mrkdwn
                  text: |-
                    *Status:* Building and passing all tests.
                    *Install URL:* ${ eas.job.expoBuildUrl }
    - eas/send_slack_message:
        if: ${ failure() }
        name: Send Slack message on failure
        inputs:
          payload:
            blocks:
              - type: header
                text:
                  type: plain_text
                  text: 'Company - :apple_logo: iOS - Develop Pipeline :red_circle:'
                  emoji: true
              - type: section
                text:
                  type: mrkdwn
                  text: |-
                    *Build Number:* ${ eas.job.version.buildNumber }
              - type: divider
              - type: section
                text:
                  type: mrkdwn
                  text: |-
                    *Status:* Build failed, see job for more information
                    *Link:* ${ eas.job.expoBuildUrl }
```

I believe that `step.shouldExecuteStep` is only about evaluating the if condition and it is trivial `success()` or `failure()` in the bug report. 
https://github.com/expo/eas-build/blob/e5d693eb29a484eda5306163df03b0369e31a39d/packages/steps/src/BuildStep.ts#L285-L317

The `try/catch` you added is a good improvement and it seems like it can really save us from erroring in some cases, but I understand it is not related to the bug report itself and it wasn't the root cause of this bug, right? We probably have some other interpolation issues when interpolating the job object.

packages/steps/src/BuildWorkflow.ts
```diff
@@ -19,7 +19,18 @@ export class BuildWorkflow {
...
+        step.ctx.logger.error({ err });
+        step.ctx.logger.error(
+          `Runner failed to evaluate if it should execute step "${step.displayName}"`
+        );
```
szdziedzic, line 29:
```suggestion
        step.ctx.logger.error(
          `Runner failed to evaluate if it should execute step "${step.displayName}", using step's if condition "${step.ifCondition}". This can be caused by trying to access non-existing object property. If you think this is a bug report it here: https://github.com/expo/eas-cli/issues.`
        );
```

packages/steps/src/BuildWorkflow.ts
szdziedzic: What config did you use to replicate the issue?

szdziedzic (approved):

packages/steps/src/BuildWorkflow.ts
radoslawkrzemien: The user who reported this said
```
however I comment out the ${ eas.job.expoBuildUrl } line as --local builds do not have a build url
```
so I took an existing config for sending slack messages and commented out a line being interpolated with non-existing value - just as the user described. And it resulted in the same problem

radoslawkrzemien: Unless we want to do some other action when the user tries to interpolate non-existing value then throwing and error

radoslawkrzemien: `shouldExecuteStep` never returns in the case described by the user, it throws an error

szdziedzic (approved):

PR #444, opened 2024-09-26
[eas-build] Expose generic job's `BuildWorkflow`

sjchmiela:
# Why

Exposing `BuildWorkflow` from `runGenericJobAsync` lets us fetch individual steps and their outputs after the job is complete.

# How

- wraps `buildWorkflow.executeAsync()` with `asyncResult` and returns the result from `runGenericJobAsync` alongside `buildWorkflow`,
- adds `outputs` to `Generic.Job` so the user has a way of telling us which steps outputs are to be considered jobs outputs,
- when a generic job finishes, Worker sends a `robotAccessToken`-authenticated request to [a new `www` endpoint](https://github.com/expo/universe/pull/16916/files).

Also changes the way we're handling `outputsDir` and `envsDir` — we want the outputs and environment variables to be collected always, not only if the step succeeded. And exposes `jsepEval` so we can use it in Worker.

Moreover, removes `outputs` from `BuildStep`. We had two: `outputs` and `outputById`, one being updated as we process the workflow and the other not.

# Test Plan

Adjusted tests. I've used this version to run a custom job with `set-output name test` and gotten 

Zrzut ekranu 2024-09-27 o 18 44 42

### DIFF
--- a/packages/steps/src/BuildStep.ts
+++ b/packages/steps/src/BuildStep.ts
@@ -120,7 +120,7 @@ export class BuildStep extends BuildStepOutputAccessor {
   public readonly displayName: string;
   public readonly supportedRuntimePlatforms?: BuildRuntimePlatform[];
   public readonly inputs?: BuildStepInput[];
-  public readonly outputs?: BuildStepOutput[];
+  public readonly outputById: BuildStepOutputById;
   public readonly command?: string;
   public readonly fn?: BuildStepFunction;
   public readonly shell: string;
@@ -133,7 +133,6 @@ export class BuildStep extends BuildStepOutputAccessor {
 
   private readonly internalId: string;
   private readonly inputById: BuildStepInputById;
-  protected readonly outputById: BuildStepOutputById;
   protected executed = false;
 
   public static getNewId(userDefinedId?: string): string {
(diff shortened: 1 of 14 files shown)


radoslawkrzemien (approved):
LGTM 🚀 

What's the rationale behind returning `runResult: Result<void>;`?

szdziedzic (approved):
looks reasonable

sjchmiela:
> What's the rationale behind returning `runResult: Result<void>;`?

Most importantly, I wanted to be able to return output always, if the job succeeds or fails. This is how it works in GitHub Actions too. Not sure if people use it, but if we can do it too… why not!

I could do `{ buildWorkflow, error: Error | null }` to denote whether the workflow succeeded or failed, or something with `type` too. In the end I figured we can use Expo's own `results` library.

packages/steps/src/BuildStep.ts
```diff
@@ -256,23 +267,22 @@ export class BuildStep extends BuildStepOutputAccessor {
...
 
-  public getOutputValueByName(name: string): string | undefined {
-    if (!this.executed) {
-      throw new BuildStepRuntimeError(
-        `Failed getting output "${name}" from step "${this.displayName}". The step has not been executed yet.`
-      );
-    }
-    if (!this.hasOutputParameter(name)) {
-      throw new BuildStepRuntimeError(`Step "${this.displayName}" does not have output "${name}".`);
+        this.ctx.logger.debug({ error }, 'Failed to collect output parameters');
```
sjchmiela, line 281:
```suggestion
        this.ctx.logger.debug({ err: error }, 'Failed to collect output parameters');
```

PR #446, opened 2024-10-11
[eas-build-job] Formalize interpolation contexts

sjchmiela:
# Why

We want to stabilize the contexts that are available to jobs and workflows.

# How

There are now:
- `WorkflowInterpolationContext` — stuff that can be used on workflow level, like job's `if`, `name`, some params
- `JobInterpolationContext` — stuff that can be used on job level, like individual steps.

Joi does not do any validation. Frankly, I suspect we don't need any validation for this — it's an internal protocol between `www` and `eas-build`.

# Test Plan

Will test in `www`.


packages/eas-build-job/src/android.ts
```diff
@@ -175,4 +178,6 @@ export const JobSchema = Joi.object({
     submitProfile: Joi.string(),
   }),
   loggerLevel: Joi.string().valid(...Object.values(LoggerLevel)),
+
+  workflowInterpolationContext: Joi.any(),
```
szdziedzic, line 182: shouldn't it be never for standard builds?

szdziedzic (approved):

packages/eas-build-job/src/android.ts
sjchmiela: Hmm, yea, I guess, for clarity we can make it so. 👍

PR #451, opened 2024-10-21
[eas-build-job] Make `initiatingUserId` and `appId` required

sjchmiela:
This is ok if we deploy in the right order:
1. `www` so it starts adding those values
2. `turtle-v2` so it can access those values.


szdziedzic (commented):
I'm not sure if it's fine, because `Job` is also used in EAS CLI to create a job object sent to WWW. I'm not sure if we want to/can set it in EAS CLI?

sjchmiela:
Yea, in `www`. Or we would just always override it in `www`.

PR #452, opened 2024-10-28
[ENG-13945][eas-build-job][build-tools] add new EAGER_BUNDLE build phase for SDK 52+

szdziedzic:
# Why

https://linear.app/expo/issue/ENG-13945/add-exportembed-eager

# How

Add a new `EAGER_BUNDLE` build phase which is only run for SDK 52+. In the build phase invoke the `npx expo export:embed --eager --platform platform` command to bundle the JS code eagerly.

Run it for both managed and generic workflow (with native directories or without them).

# Test Plan

Test locally:

### iOS generic:

> [!NOTE]
> The way we display the new build phase on the website will change and use the display name of the new build phase once we update `@expo/eas-build-job` there.

### DIFF
--- a/packages/build-tools/src/builders/android.ts
+++ b/packages/build-tools/src/builders/android.ts
@@ -106,15 +106,20 @@ async function buildAsync(ctx: BuildContext<Android.Job>): Promise<void> {
 
   if (ctx.metadata?.sdkVersion && semver.satisfies(ctx.metadata?.sdkVersion, '>=52')) {
     await ctx.runBuildPhase(BuildPhase.EAGER_BUNDLE, async () => {
-      await eagerBundleAsync(ctx, {
-        ...(resolvedExpoUpdatesRuntimeVersion?.runtimeVersion
-          ? {
-              extraEnv: {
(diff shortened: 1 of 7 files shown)


szdziedzic:
* **#452** <a href="https://app.graphite.dev/github/pr/expo/eas-build/452?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @szdziedzic and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

sjchmiela (approved):
Looks reasonable to me.

I would maybe, instead of creating an `eagerBundleAsync`, inline the `runExpoCliCommandAsync`. Or, if I were to do `eagerBundleAsync` I would probably name the `runtimeVersionOverride: string | null` parameter explicitly or something like that.

After thinking about it, maybe `platform` parameter should be explicit too? Then we could offer it as a build function which developers could reuse for building multiple apps in different environments?

Looks good though.

PR #459, opened 2024-10-30
[build-tools] Log `stdout` if `build:internal` fails

sjchmiela:
# Why

I have trouble understanding why GitHub builds on the new Orchestrator fail consistently.

# How

I want to know more about the error than "GraphQL request failed". So, I want to print the second part of the output, the stdout.

# Test Plan

I ran a local build with overridden `EXPO_TOKEN=invalid`, and printing both stdout and stderr and got:

Zrzut ekranu 2024-10-30 o 22 26 23

### DIFF
--- a/packages/build-tools/src/common/easBuildInternal.ts
+++ b/packages/build-tools/src/common/easBuildInternal.ts
@@ -44,29 +44,42 @@ export async function runEasBuildInternalAsync<TJob extends BuildJob>({
     autoSubmitArgs.push('--auto-submit');
   }
 
-  const result = await spawn(
-    cmd,
-    [
-      ...args,
-      'build:internal',
-      '--platform',
-      job.platform,
-      '--profile',
-      buildProfile,
-      ...autoSubmitArgs,
-    ],
-    {
-      cwd,
-      env: {
-        ...env,
-        EXPO_TOKEN: nullthrows(job.secrets, 'Secrets must be defined for non-custom builds')
-          .robotAccessToken,
-        ...extraEnv,
-      },
-      logger,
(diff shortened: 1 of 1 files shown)


packages/build-tools/src/common/easBuildInternal.ts
```diff
@@ -44,29 +44,42 @@ export async function runEasBuildInternalAsync<TJob extends BuildJob>({
...
+      }
+    );
+  } catch (err: any) {
+    // Logging of build:internal is weird. We set mode = STDERR_ONLY_AS_STDOUT,
+    // because we don't want to pipe stdout to to the logger as it will contain
+    // the job object with its secrets. However, if the build fails, stderr alone
+    // is often not enough to debug the issue, so we also log stdout in those cases.
+    // It will look awkward (first stderr from pipe, then stdout from here), but
+    // it's better than nothing.
+    logger.error(`${err.stdout}`);
```
szdziedzic, line 79: I assume that printing job object with it's secrets is the last operation performed by this command, right? If so if there is any error there shouldn't be any secrets in stdout?

szdziedzic (approved):

packages/build-tools/src/common/easBuildInternal.ts
sjchmiela: It's not last per se, but it's close to last. After we print the job we may also start a submission (which theoretically may fail).

Maybe we should revert this after we find the issue with the orchestrator?

szdziedzic: Yeah, we can if it contains any sensitive data and there is a risk of leaking it 🤔

PR #460, opened 2024-11-01
[updates] Change fingerprint diff to use versioned fingerprint CLI

wschurman:
# Why

This adds a call into new @expo/fingerprint CLI (https://github.com/expo/expo/pull/32541) to do diffing when possible, falling back to the local diff when unavailable.

This was done to avoid needing to backport changes from the versioned package in the future to test changes in build.

# Test Plan

Test project with versioned fingerprint CLI:

1. Create beta project
2. `eas update:configure`
3. Change app config to app.config.js and add `"version": process.env.EAS_BUILD ? "1.0.0" : "2.0.0",`
3. Add console.log to CLI resolution to verify local CLI is resolved and used.
4. `~/expo/run-with-local-eas-build.sh build --local  --build-logger-level=debug`, see no log indicating that it fell back to local diff indicating that it is using the local copy in the project to do the diffing. See diff looks correct (just config due to process.env modification above).

Test project without (same steps as above but with a SDK 51 project). See it debug logs that it is falling back to local.

### DIFF
--- a/packages/build-tools/src/utils/expoFingerprintCli.ts
+++ b/packages/build-tools/src/utils/expoFingerprintCli.ts
@@ -1,0 +1,84 @@
+import resolveFrom from 'resolve-from';
+import spawnAsync from '@expo/turtle-spawn';
+import { BuildStepEnv } from '@expo/steps';
+import fs from 'fs-extra';
+import semver from 'semver';
+
+export class ExpoFingerprintCLIModuleNotFoundError extends Error {}
+export class ExpoFingerprintCLIInvalidCommandError extends Error {}
+export class ExpoFingerprintCLICommandFailedError extends Error {}
+
+function resolveExpoFingerprintCLI(projectRoot: string): string {
+  const expoPackageRoot = resolveFrom.silent(projectRoot, 'expo/package.json');
+  try {
+    return (
+      resolveFrom.silent(expoPackageRoot ?? projectRoot, '@expo/fingerprint/bin/cli') ??
+      resolveFrom(expoPackageRoot ?? projectRoot, '@expo/fingerprint/bin/cli.js')
+    );
+  } catch (e: any) {
+    if (e.code === 'MODULE_NOT_FOUND') {
+      throw new ExpoFingerprintCLIModuleNotFoundError(
+        `The \`@expo/fingerprint\` package was not found.`
+      );
+    }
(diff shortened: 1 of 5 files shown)


wschurman:
* **#460** <a href="https://app.graphite.dev/github/pr/expo/eas-build/460?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @wschurman and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/build-tools/src/utils/expoFingerprintCli.ts
```diff
@@ -0,0 +1,47 @@
...
+export async function expoFingerprintCommandAsync(
+  projectDir: string,
+  args: string[],
+  { env }: { env: BuildStepEnv }
+): Promise<string> {
+  let expoFingerprintCli;
+  try {
+    expoFingerprintCli =
+      silentResolveFrom(projectDir, '@expo/fingerprint/bin/cli') ??
+      resolveFrom(projectDir, '@expo/fingerprint/bin/cli.js');
```
Kudo, line 18: we should not directly resolve `@expo/fingerprint` from projectDir. that may break from pnpm isolated mode. we should resolve `projectDir -> expo -> `@expo/fingerprint`, or even re-export `expo/fingerprint-cli`

example for `projectDir -> expo -> `@expo/fingerprint` resolver
https://github.com/expo/expo/blob/ca42e1312eaf3238212ec437f7fa9583db78ea0c/packages/%40expo/fingerprint/src/ExpoVersions.ts#L20-L24

Kudo (changes requested):

packages/build-tools/src/utils/expoFingerprintCli.ts
wschurman: Good call. I'll wait for a canary to be published with @quinlanj's change (making fingerprint a dependency of expo) since testing this is fairly complex otherwise.

Kudo (approved):

wschurman:
### Merge activity

* **Nov 7, 1:17 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/expo/eas-build/460).

PR #463, opened 2024-11-12
[build-tools] use frozen lockfile for node modules install for SDK 52+ and RN 0.76+

szdziedzic:
# Why

https://exponent-internal.slack.com/archives/C02123T524U/p1731420965266849?thread_ts=1731420501.834709&cid=C02123T524U
https://exponent-internal.slack.com/archives/C02123T524U/p1731421835879399

We want to use frozen lockfile when installing `node_modules` on our CI.

# How

Enable using frozen lockfile for SDK 52+ and RN 0.76+ in the `install node modules` phase.

Check for `EAS_NO_FROZEN_LOCKFILE` env var to skip using frozen lockfile if set.

Don't use frozen lockfile after prebuild as prebuild can sometimes modify `the package.json`.

# Test Plan

Test locally

### DIFF
--- a/packages/build-tools/src/common/installDependencies.ts
+++ b/packages/build-tools/src/common/installDependencies.ts
@@ -18,7 +18,7 @@ export async function installDependenciesAsync({
   packageManager: PackageManager;
   env: Record<string, string | undefined>;
   cwd: string;
-  logger?: SpawnOptions['logger'];
+  logger: Exclude<SpawnOptions['logger'], undefined>;
   infoCallbackFn?: SpawnOptions['infoCallbackFn'];
   useFrozenLockfile: boolean;
 }): Promise<{ spawnPromise: SpawnPromise<SpawnResult> }> {
@@ -50,7 +50,7 @@ export async function installDependenciesAsync({
   if (env['EAS_VERBOSE'] === '1') {
     args = [...args, '--verbose'];
   }
-  logger?.info(`Running "${packageManager} ${args.join(' ')}" in ${cwd} directory`);
+  logger.info(`Running "${packageManager} ${args.join(' ')}" in ${cwd} directory`);
   return {
     spawnPromise: spawn(packageManager, args, {
       cwd,
--- a/packages/build-tools/src/common/installDependencies.ts
+++ b/packages/build-tools/src/common/installDependencies.ts
@@ -1,6 +1,5 @@
 import path from 'path';
 
-import semver from 'semver';
 import { Job } from '@expo/eas-build-job';
 import spawn, { SpawnPromise, SpawnResult, SpawnOptions } from '@expo/turtle-spawn';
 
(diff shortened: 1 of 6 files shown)


szdziedzic:
* **#463** <a href="https://app.graphite.dev/github/pr/expo/eas-build/463?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @szdziedzic and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/build-tools/src/steps/functions/installNodeModules.ts
```diff
@@ -44,7 +44,7 @@ export async function installNodeModules(
   if (packageManager === PackageManager.PNPM) {
     args = ['install', '--no-frozen-lockfile'];
```
Kudo, line 45: question: do we need add frozen-lockfile in this installNodeModules.ts?

packages/build-tools/src/common/installDependencies.ts
```diff
@@ -1,24 +1,61 @@
...
-    const isYarn2 = await isUsingYarn2(ctx.getReactNativeProjectDirectory());
-    if (isYarn2) {
-      args = ['install', '--no-immutable', '--inline-builds'];
+  const shouldUseFrozenLockfile = Boolean(
+    !withoutFrozenLockfile &&
+      !ctx.env.EAS_NO_FROZEN_LOCKFILE &&
+      ((!!sdkVersionFromPackageJson && semver.satisfies(sdkVersionFromPackageJson, '>=52')) ||
+        (!!reactNativeVersionFromPackageJson &&
+          semver.satisfies(reactNativeVersionFromPackageJson, '>=0.76')))
+  );
```
Kudo, line 35: the logic looks complicated. maybe refactor some logic out to a function
```ts
  const shouldUseFrozenLockfile = Boolean(
    !withoutFrozenLockfile &&
    !ctx.env.EAS_NO_FROZEN_LOCKFILE) ?? useDefaultFrozenLockfile(packageJson);
    
  function useDefaultFrozenLockfile(packageJson: string): boolean {
    const sdkVersion = semver.coerce(packageJson?.dependencies?.expo)?.version;
    const reactNativeVersion = semver.coerce(
      packageJson?.dependencies?.['react-native']
    )?.version;
    if (!sdkVersion || sdkVersion.major < 52) {
      return false;
    }
    if (!reactNativeVersion || reactNativeVersion.major < 76) {
      return false;
    }
    return true;
  }
```

passing packageJson to the function and we can also save a little time without doing semver parsing when `EAS_NO_FROZEN_LOCKFILE=true`.

sjchmiela:
This pull request now also changes the default to "DO use frozen lockfile". This will kick in if we can't prove the versions requested by the user in `package.json` are `expo<52` or `react-native<0.76`. I'm a bit afraid of this change, but I also don't know how would we go about making this breaking change in a nice-to-users way. Do you have any ideas?

PR #464, opened 2024-11-13
[build-tools] bump @expo/config and @expo/config-plugins version

gabrieldonadel:
# Why

Need to bump `@expo/config` for the SDK 52 release in order to read the app config `newArchEnabled` value in turtle-v2 

# How

bump `@expo/config` and `@expo/config-plugins` version

# Test Plan

CI should be green

### DIFF
--- a/packages/build-tools/package.json
+++ b/packages/build-tools/package.json
@@ -22,8 +22,8 @@
   "bugs": "https://github.com/expo/eas-build/issues",
   "license": "BUSL-1.1",
   "dependencies": {
-    "@expo/config": "9.0.3",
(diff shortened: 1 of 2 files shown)


szdziedzic (commented):
I believe that to make it work in all cases after this upgrade, we would optimally need to do sth as I did for EAS CLI recently: https://github.com/expo/eas-cli/pull/2529 (use `npx expo config` call to resolve config first if available to be able to resolve versioned config). I can try to do the same change here and bump the dependencies. Will let you know when it is ready.

sjchmiela:
Done in https://github.com/expo/eas-build/pull/475

PR #470, opened 2024-11-21
[build-tools] [ENG-13708] Upload outputs in public step

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-13708/move-uploading-outputs-to-eas-build-and-public-step

# How

Created outputs upload function based on the one existing in turtle, but accepting the `logger` and `expoApiV2BaseUrl` params.
Calling that function at the end of generic job in a new dedicated `Complete job` step, with the logger for that step.
Added mappings for that new step.

# Test Plan

Automated tests mostly based on those from Turtle

# Deployment plan
1) Merge this PR
2) Release new version of `build-tools` and `eas-build-job`
3) Update `eas-build-job` in `website`, merge and deploy
4) Update `build-tools` and `eas-build-job` in Turtle, merge and deploy

### DIFF
--- a/packages/build-tools/src/utils/outputs.ts
+++ b/packages/build-tools/src/utils/outputs.ts
@@ -56,7 +56,7 @@ export function getJobOutputsFromSteps({
 }: {
   jobOutputDefinitions: Record<string, string>;
   interpolationContext: {
-    steps: Record<string, { outputs: Record<string, string | undefined> } | string>;
+    steps: Record<string, { outputs: Record<string, string | undefined> }>;
   };
 }): Record<string, string | undefined> {
   const jobOutputs: Record<string, string | undefined> = {};
--- a/packages/build-tools/src/utils/outputs.ts
+++ b/packages/build-tools/src/utils/outputs.ts
@@ -51,6 +51,7 @@ export async function uploadJobOutputsToWwwAsync(
     }
   } catch (err) {
     logger.error({ err }, 'Failed to upload outputs');
+    throw err;
   }
 }
 
--- a/packages/build-tools/src/utils/outputs.ts
+++ b/packages/build-tools/src/utils/outputs.ts
@@ -25,9 +25,9 @@ export async function uploadJobOutputsToWwwAsync(
     const robotAccessToken = nullthrows(ctx.job.secrets?.robotAccessToken);
 
     const interpolationContext = {
-      steps: Object.fromEntries(steps.map((step) => [step.id, getDynamicValuesFromStep(step)])),
+      steps: Object.fromEntries(steps.map((step) => [step.id, getStepOutputsAsObject(step)])),
     };
-    logger.info({ dynamicValues: interpolationContext }, 'Using dynamic values');
+    logger.debug({ dynamicValues: interpolationContext }, 'Using dynamic values');
 
     const outputs = getJobOutputsFromSteps({
       jobOutputDefinitions: ctx.job.outputs,
@@ -55,7 +55,9 @@ export function getJobOutputsFromSteps({
   interpolationContext,
 }: {
   jobOutputDefinitions: Record<string, string>;
-  interpolationContext: Record<string, any>;
+  interpolationContext: {
+    steps: Record<string, { outputs: Record<string, string | undefined> } | string>;
+  };
 }): Record<string, string | undefined> {
   const jobOutputs: Record<string, string | undefined> = {};
   for (const [outputKey, outputDefinition] of Object.entries(jobOutputDefinitions)) {
@@ -70,7 +72,7 @@ export function getJobOutputsFromSteps({
 }
 
 /** This is what we'll use to generate an object representing a step. */
-export function getDynamicValuesFromStep(step: BuildStep): {
+export function getStepOutputsAsObject(step: BuildStep): {
   outputs: Record<string, string | undefined>;
 } {
   const outputs = Object.fromEntries(
--- a/packages/build-tools/src/utils/outputs.ts
+++ b/packages/build-tools/src/utils/outputs.ts
@@ -2,9 +2,10 @@ import { Generic } from '@expo/eas-build-job';
 import { BuildStep, jsepEval } from '@expo/steps';
 import { bunyan } from '@expo/logger';
 import nullthrows from 'nullthrows';
-import fetch from 'node-fetch';
 
 import { BuildContext } from '../context';
+
+import { turtleFetch } from './turtleFetch';
 
 export async function uploadJobOutputsToWwwAsync(
   ctx: BuildContext<Generic.Job>,
@@ -34,21 +35,14 @@ export async function uploadJobOutputsToWwwAsync(
     });
     logger.info('Uploading outputs');
 
-    const response = await fetch(
-      new URL(`workflows/${workflowJobId}`, expoApiV2BaseUrl).toString(),
-      {
-        method: 'PATCH',
-        body: JSON.stringify({ outputs }),
-        headers: {
-          Authorization: `Bearer ${robotAccessToken}`,
-          'Content-Type': 'application/json',
-        },
-        timeout: 20000,
-      }
-    );
-    if (!response.ok) {
-      throw new Error(`[${response.status}] ${response.statusText}`);
-    }
+    await turtleFetch(new URL(`workflows/${workflowJobId}`, expoApiV2BaseUrl).toString(), 'PATCH', {
+      json: { outputs },
+      headers: {
+        Authorization: `Bearer ${robotAccessToken}`,
+      },
+      timeout: 20000,
+      logger,
+    });
   } catch (err) {
     logger.error({ err }, 'Failed to upload outputs');
     throw err;
--- a/packages/build-tools/src/utils/outputs.ts
+++ b/packages/build-tools/src/utils/outputs.ts
@@ -34,15 +34,21 @@ export async function uploadJobOutputsToWwwAsync(
     });
     logger.info('Uploading outputs');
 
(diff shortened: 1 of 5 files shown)


khamilowicz (approved):
LGTM, it would be nice to see a PR with updates to turtle.

radoslawkrzemien:
@khamilowicz https://github.com/expo/turtle-v2/pull/1993

packages/build-tools/src/utils/outputs.ts
```diff
@@ -0,0 +1,80 @@
...
+    expoApiV2BaseUrl,
+  }: { steps: BuildStep[]; logger: bunyan; expoApiV2BaseUrl: string }
+): Promise<void> {
+  if (!ctx.job.outputs) {
+    logger.info('Job defines no outputs, skipping upload');
+    return;
+  }
+
+  try {
+    const workflowJobId = nullthrows(ctx.job.builderEnvironment?.env?.__WORKFLOW_JOB_ID);
```
szdziedzic, line 23: If the user maliciously changes the `__WORKFLOW_JOB_ID` value to some other workflow ID they won't be able to write outputs values for it because we validate it in WWW, right?

BTW It would be good to move it to `job.workflowId` so it can't be overridden by the user (not a part of this PR)

packages/build-tools/src/utils/outputs.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    await fetch(new URL(`workflows/${workflowJobId}`, expoApiV2BaseUrl).toString(), {
+      method: 'PATCH',
+      body: JSON.stringify({ outputs }),
+      headers: {
+        Authorization: `Bearer ${robotAccessToken}`,
+        'Content-Type': 'application/json',
+      },
+      timeout: 20000,
+    });
```
szdziedzic, line 45: I believe that `fetch` won't throw if the status code is for example 500 by default. You need to check it manually. It will only throw on things like timeouts.
```suggestion
    const response = await fetch(new URL(`workflows/${workflowJobId}`, expoApiV2BaseUrl).toString(), {
      method: 'PATCH',
      body: JSON.stringify({ outputs }),
      headers: {
        Authorization: `Bearer ${robotAccessToken}`,
        'Content-Type': 'application/json',
      },
      timeout: 20000,
    });
   if (!response.ok) {
     throw new Error(...)
   }
```

packages/build-tools/src/utils/outputs.ts
```diff
@@ -0,0 +1,80 @@
...
+  } catch (err) {
+    logger.error({ err }, 'Failed to upload outputs');
```
szdziedzic, line 47: Shouldn't we fail the workflow run altogether when an error is thrown? If you expect outputs from this workflow in your other runs I guess it would be hard to debug later on if we mark this run as a `success` but some required inputs will be missing.

What we can do is retry outputs upload a couple of times and if it still fails throw an error and mark workflow as errored.

packages/build-tools/src/utils/outputs.ts
```diff
@@ -0,0 +1,80 @@
...
+    };
+    logger.info({ dynamicValues: interpolationContext }, 'Using dynamic values');
+
+    const outputs = getJobOutputsFromSteps({
+      jobOutputDefinitions: ctx.job.outputs,
+      interpolationContext,
+    });
+    logger.info('Uploading outputs');
+
+    await fetch(new URL(`workflows/${workflowJobId}`, expoApiV2BaseUrl).toString(), {
```
szdziedzic, line 37: do we want to add retries for it? I believe they would be useful

szdziedzic, line 37: we can add a util like `turtleFetch` here if we want to:
https://github.com/expo/turtle-v2/blob/998063fdfb9edb61eba3e080a960a0de067139ab/src/libs/turtle-common/src/turtleFetch.ts#L16-L74

szdziedzic, line 37: the WWW endpoint is already working on prod, right?

packages/build-tools/src/utils/outputs.ts
```diff
@@ -0,0 +1,80 @@
...
+  jobOutputDefinitions: Record<string, string>;
+  interpolationContext: Record<string, any>;
+}): Record<string, string | undefined> {
+  const jobOutputs: Record<string, string | undefined> = {};
+  for (const [outputKey, outputDefinition] of Object.entries(jobOutputDefinitions)) {
+    const outputValue = outputDefinition.replace(/\$\{\{(.+?)\}\}/g, (_match, expression) => {
+      return `${jsepEval(expression, interpolationContext)}`;
+    });
+
+    jobOutputs[outputKey] = outputValue;
```
szdziedzic, line 65: the `jobOutputs` can be only of type string right?

packages/build-tools/src/utils/outputs.ts
radoslawkrzemien: It *used to* use `turtleFetch` when it was located in turtle, but it's unavailable here so I replace it with the standard `fetch`, but you're right that we need something similar, if only just for retries

packages/build-tools/src/utils/outputs.ts
radoslawkrzemien: https://linear.app/expo/issue/ENG-14365/move-builderenvironmentenv-workflow-id-to-jobworkflowid

packages/build-tools/src/utils/outputs.ts
radoslawkrzemien: Record<string, string | undefined>

packages/build-tools/src/utils/outputs.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  return jobOutputs;
+}
+
+/** This is what we'll use to generate an object representing a step. */
+export function getDynamicValuesFromStep(step: BuildStep): {
```
szdziedzic, line 73: nit
```suggestion
export function getStepOutputsAsObject(step: BuildStep): {
```

WDYT?

packages/build-tools/src/utils/outputs.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+
+/** Function we use to get outputs of the whole job from steps. */
+export function getJobOutputsFromSteps({
+  jobOutputDefinitions,
+  interpolationContext,
+}: {
+  jobOutputDefinitions: Record<string, string>;
+  interpolationContext: Record<string, any>;
```
szdziedzic, line 58: don't use `any`
```suggestion
  interpolationContext: Record<string, Record<string, string | undefined>>;
```

szdziedzic, line 58: this is always created only based on steps outputs

is there a chance that job output definition can be based on some other context type than step output? Like generic job context with `ctx.platform` and so on?

packages/build-tools/src/utils/outputs.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  try {
+    const workflowJobId = nullthrows(ctx.job.builderEnvironment?.env?.__WORKFLOW_JOB_ID);
+    const robotAccessToken = nullthrows(ctx.job.secrets?.robotAccessToken);
+
+    const interpolationContext = {
+      steps: Object.fromEntries(steps.map((step) => [step.id, getDynamicValuesFromStep(step)])),
+    };
+    logger.info({ dynamicValues: interpolationContext }, 'Using dynamic values');
```
szdziedzic, line 30: I believe that parts of the logs that are not part of the `msg` field are not shown on the website.
```suggestion
    logger.info(`Using dynamic values: ${JSON.stringify(interpolationContext)}`);
```

szdziedzic (changes requested):

packages/build-tools/src/utils/outputs.ts
radoslawkrzemien: AFAIK no, just step outputs

packages/build-tools/src/utils/outputs.ts
radoslawkrzemien: You are right that it is not visible. However, now that I think on it, I'm not sure if it should be printed to the user. It's been copied from the existing implementation in Turtle, which didn't print anything to the user, but this particular line looks to me as better suited to be printed in the non-user-facing logs only, wdyt?

szdziedzic: yeah, we can add it to non-user-facing logs

packages/build-tools/src/utils/outputs.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/** Function we use to get outputs of the whole job from steps. */
+export function getJobOutputsFromSteps({
+  jobOutputDefinitions,
+  interpolationContext,
+}: {
+  jobOutputDefinitions: Record<string, string>;
+  interpolationContext: {
+    steps: Record<string, { outputs: Record<string, string | undefined> } | string>;
```
szdziedzic, line 59: I'm curious why `| string` is needed here?

szdziedzic: Aren't all steps object of type
```ts
Record<string, { outputs: Record<string, string | undefined> }>
```
?

radoslawkrzemien: They are, but the type check complains. I'm also not sure why 🤷

radoslawkrzemien: Actually `yarn tsc` does not complain, just my IDE. Removed the `| string`

radoslawkrzemien: I *think* it's because `Object.fromEntries` gets a `[string, Record<string, string | undefined>][]` param

PR #473, opened 2024-12-03
[build-tools] Remove fallback unversioned fingerprint diff

wschurman:
# Why

Follow-up to https://github.com/expo/eas-build/pull/460/.

Since the output of the versioned diff fingerprint CLI is itself versioned (can change between fingerprint versions), we should no longer show the diff if it can't be computed in a versioned manner since the hardcoded local types may not be correct for all versions.

In theory most projects should now have a versioned fingerprint CLI (as of 52).

Closes ENG-14388.

# How

Remove fallback code.

# Test Plan

`tsc`

(this removes code and the tests for that removed code)


wschurman:
* **#473** <a href="https://app.graphite.dev/github/pr/expo/eas-build/473?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-build/473?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

Kudo:
> Since the output of the versioned diff fingerprint CLI is itself versioned (can change between fingerprint versions)

not having enough knowledge in eas build, i could miss the context here. could you share where the relevant code for the versioned output? especially for sdk 51 or lower projects, how do they diff fingerprint?

wschurman:
> could you share where the relevant code for the versioned output?

The `fingerprint:diff` CLI command from the fingerprint package (as well as the diffFingerprints function) outputs a structure of the form `FingerprintDiffItem`, which is a structure that is defined in the versioned package. Thus, it is subject to change between SDK versions ([it changed from SDK 51 to 52, for example](https://github.com/expo/expo/commit/5512083856b21cff467b012d6cf5e3bc3c12b02d)). Rather than lock us into this format forever, it's preferable to only operate on this format in a versioned context. Thus, we only do the diff and parse the results in the versioned CLI.

> especially for sdk 51 or lower projects, how do they diff fingerprint?

They don't. Fingerprint runtime version is still in beta so we don't need to support it in perpetuity, so now is as good a time as any to remove support to reduce attention cost of a beta feature.

quinlanj (approved):

Kudo:
>> especially for sdk 51 or lower projects, how do they diff fingerprint?

> They don't. Fingerprint runtime version is still in beta so we don't need to support it in perpetuity, so now is as good a time as any to remove support to reduce attention cost of a beta feature.

oh cool, that answers my question and is fair. thanks!

Kudo (approved):

wschurman:
### Merge activity

* **Dec 4, 1:01 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/expo/eas-build/473).

PR #479, opened 2024-12-17
Add `environment` to job schemas

khamilowicz:
# Why

[ENG-14307: move environment to job object from metadata](https://linear.app/expo/issue/ENG-14307/move-environment-to-job-object-from-metadata)

Allow storing `environment` on the `job` object.


packages/eas-build-job/src/android.ts
```diff
@@ -183,4 +183,6 @@ export const JobSchema = Joi.object({
 
   initiatingUserId: Joi.string().required(),
   appId: Joi.string().required(),
+
+  environment: Joi.string().valid('production', 'preview', 'development'),
```
szdziedzic, line 187: I believe we also need to update typescript `Job` interface

szdziedzic (changes requested):

szdziedzic (approved):

PR #480, opened 2024-12-17
[eas-cli] allow running eas/build function group with credentials for iOS GH builds

szdziedzic:
# Why

We are making a decision in our system whether to execute the credentials or without the credentials flow for the iOS `eas/build` function group based on the presence of `job.secrets.buildCredentials`. However, at the time of making this discussion (before the build starts), these are always missing for GH builds and are only resolved later.

We should rely on `job.simulator` only instead, which is always correctly set for GH builds https://github.com/expo/universe/blob/9eccb36e30da89e17d1c8b748437da99faebfa7d/server/www/src/data/entities/github/GitHubBuildUtils.ts#L165

# How

Only use `job.simulator` to decide which flow to enter

# Test Plan

Test it manually

### DIFF
--- a/packages/build-tools/src/steps/functionGroups/build.ts
+++ b/packages/build-tools/src/steps/functionGroups/build.ts
@@ -36,7 +36,7 @@ export function createEasBuildBuildFunctionGroup(
     createBuildStepsFromFunctionGroupCall: (globalCtx) => {
       if (buildToolsContext.job.platform === Platform.IOS) {
         // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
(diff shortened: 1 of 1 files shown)


szdziedzic:
* **#480** <a href="https://app.graphite.dev/github/pr/expo/eas-build/480?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-build/480?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sjchmiela (approved):

PR #484, opened 2025-01-09
[steps] Resolve absolute `working_directory` paths from repository root

sjchmiela:
# Why

With Workflows I expect more and more users will want to execute scripts related less to mobile app, in directories other than that of the mobile app.

Currently, `working_directory` works like this:
- relative paths get resolved from the base directory (mobile app directory) — `working_directory: scripts` if base directory is `apps/mobile` ends up being `/home/expo/build/apps/mobile/scripts`
- absolute paths get resolved from VM root — `working_directory: /usr/local/bin` ends up being `/usr/local/bin`.

This makes it cumbersome to execute scripts in directories other than that of mobile app. Say, a user wants to execute scripts in `/apps/web`. If base directory is `apps/mobile`, they'd need to use `../web`. This makes `working_directory` coupled to base directory set in GitHub repository settings.

# How

Here I'm special-casing absolute `working_directory` paths. I expect very few people to need to execute scripts in FS-absolute paths (`/usr/local/bin`). So now it is going to work like this:
- (unchanged) relative paths get resolved from the base directory (mobile app directory) — `working_directory: scripts` if base directory is `apps/mobile` ends up being `/home/expo/build/apps/mobile/scripts`
- absolute paths get resolved from repository root — `working_directory: /usr/local/bin` ends up being `/home/expo/build/usr/local/bin`.

# Test Plan

Adjusted tests.

Also tested manually. A workflow like

```yaml
jobs:
  test:
    steps:
      - run: pwd
      - run: pwd
        working_directory: /
      - run: pwd
        working_directory: ./
      - uses: eas/checkout
      - run: pwd
        working_directory: scripts
      - run: pwd
        working_directory: /apps/web
```

produces output like this
Zrzut ekranu 2025-01-9 o 14 12 39

### DIFF
--- a/packages/steps/src/BuildStepContext.ts
+++ b/packages/steps/src/BuildStepContext.ts
@@ -233,9 +233,15 @@ export class BuildStepContext {
   }
 
   public get workingDirectory(): string {
-    return this.relativeWorkingDirectory
-      ? path.resolve(this.ctx.defaultWorkingDirectory, this.relativeWorkingDirectory)
-      : this.ctx.defaultWorkingDirectory;
+    if (!this.relativeWorkingDirectory) {
+      return this.ctx.defaultWorkingDirectory;
+    }
+
+    if (path.isAbsolute(this.relativeWorkingDirectory)) {
+      return path.join(this.ctx.projectTargetDirectory, this.relativeWorkingDirectory);
+    }
+
+    return path.join(this.ctx.defaultWorkingDirectory, this.relativeWorkingDirectory);
   }
 
   public serialize(): SerializedBuildStepContext {
--- a/packages/steps/src/__tests__/BuildConfigParser-test.ts
+++ b/packages/steps/src/__tests__/BuildConfigParser-test.ts
@@ -179,7 +179,9 @@ describe(BuildConfigParser, () => {
       expect(step5.id).toMatch(UUID_REGEX);
       expect(step5.name).toBe('List files in another directory');
       expect(step5.command).toBe('ls -la');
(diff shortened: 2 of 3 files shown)


packages/steps/src/BuildStepContext.ts
```diff
@@ -233,9 +233,15 @@ export class BuildStepContext {
...
 
   public get workingDirectory(): string {
-    return this.relativeWorkingDirectory
-      ? path.resolve(this.ctx.defaultWorkingDirectory, this.relativeWorkingDirectory)
-      : this.ctx.defaultWorkingDirectory;
+    if (!this.relativeWorkingDirectory) {
+      return this.ctx.defaultWorkingDirectory;
+    }
+
+    if (path.isAbsolute(this.relativeWorkingDirectory)) {
```
szdziedzic, line 240: nit: maybe we should rename `relativeWorkingDirectory` to just `workingDirectoryInput` or something like this since it might not be relative

szdziedzic (approved):
We discussed it offline and it looks reasonable overall. 

Do you think this change can break people's existing workflows? Do you think anybody relied on `working_dir: /absolute/path` at the moment that it would become `/project/target/dir/absolute/path`? It would be a breaking change then. We technically always recommended using it as a relative path, AFAIK, so maybe that's not the case.

We probably also need to update docs as a follow-up.

packages/steps/src/BuildStepContext.ts
sjchmiela: Yea, I wish all `steps` were significantly simpler. When we move `worker` in we may think of gradually simplifying all this.

PR #485, opened 2025-01-09
[build-tools] run pre-install hook with npm when yarn is detected as package manager

szdziedzic:
# Why

https://exponent-internal.slack.com/archives/C9PRD479V/p1736426668589209

# How

Always run the pre-install hook with `npm run` when `yarn` is detected as a package manager to avoid issues with corepack


byCedric (approved):
Thank you!

PR #486, opened 2025-01-10
[build-tools] don't run `EAGER_BUNDLE` build phase for dev client builds

szdziedzic:
# Why

https://exponent-internal.slack.com/archives/C017N0N99RA/p1736460082340769

# How

If `ctx.metadata?.developmentClient` is true, skip running the phase

# Test Plan

Tests


sjchmiela (approved):
I think we should also update `functionGroups/build`?

PR #490, opened 2025-01-13
[steps] Allow all values to `steps.with`

sjchmiela:
# Why

Currently, it seems the union fails to accept booleans, because `z.boolean()` is outside of the array literal.

# How

Replaced explicit schema with `z.unknown()`, thus letting the build step input parser validate the params (which afaik can be `json`, so anything `JSON.parse` can output).

After this is merged, going to update `eas-build-job` in `www` and change `with` definition in `WorkflowCustomJob`.

---

I also decided to replace some more "type assumptions" to `unknown`. So far our types were often very optimistic, e.g. that a build step input (the one we get from the user) will only ever be of the right type. I've changed that to `unknown` and more type changes needed to follow. Fortunately, those are only type changes (there is [_one_ code change](https://github.com/expo/eas-build/pull/490/files#diff-5dce43c6bc6a8d5435521a29218e8a4d3f70c68c8c73bd88083857a6e5d334e9R110)).

# Test Plan

Adjusted tests.

### DIFF
--- a/packages/eas-build-job/src/__tests__/step.test.ts
+++ b/packages/eas-build-job/src/__tests__/step.test.ts
@@ -33,6 +33,8 @@ describe('StepZ', () => {
           key2: ['value1'],
         },
         arg4: '${{ steps.step1.outputs.test }}',
+        arg5: true,
+        arg6: [1, 2, 3],
       },
     };
     expect(StepZ.parse(step)).toEqual(step);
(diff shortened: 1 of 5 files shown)


szdziedzic (changes requested):
Tests are broken

szdziedzic (approved):

PR #491, opened 2025-01-13
[steps] [ENG-14503] Don't wrap b64-encoded values on Linux

radoslawkrzemien:
# Why

Linux systems add line wraps to b64 encoded values after 76 characters by default. We don't want that

# How

If `set-env` is executed on a Linux system `base64` command is called with `-w 0` option. It's not present in other systems therefore in other case `base64` is called with no options

# Test Plan

Automated tests pass
Tested manually both for `ios` and `android`


packages/steps/bin/set-env
```diff
@@ -20,4 +20,8 @@ if [[ "$NAME" == *"="* ]]; then
   exit 1
 fi
 
-echo -n "$VALUE" | base64 > $__EXPO_STEPS_ENVS_DIR/$NAME
+if [[ "$OSTYPE" == "linux"* ]]; then
+  echo -n "$VALUE" | base64 -w 0 > $__EXPO_STEPS_ENVS_DIR/$NAME
+else
+  echo -n "$VALUE" | base64 > $__EXPO_STEPS_ENVS_DIR/$NAME
+fi
```
sjchmiela, line 27:
```suggestion
echo -n "$VALUE" | base64 -w 0 > $__EXPO_STEPS_ENVS_DIR/$NAME
```
`-w [number]` works in macOS (even though `--help` doesn't mention it). Can you confirm it works in Linux too?

sjchmiela (changes requested):

packages/steps/bin/set-env
radoslawkrzemien: I can confirm it doesn't work on macOS, got an error trying to run it on my machine

radoslawkrzemien:
```
radoslawkrzemien@MacBook-Pro-9 ~ % echo czescstaszekawsumiedobrze | base64 -w 0
base64: invalid option -- w
Usage:	base64 [-Ddh] [-b num] [-i in_file] [-o out_file]
  -b, --break    break encoded string into num character lines
  -Dd, --decode   decodes input
  -h, --help     display this message
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
```

sjchmiela: Fascinating! On my computer it doesn't do that!

```
> echo czescradekcotamuciebie | base64 -w 0
Y3plc2NyYWRla2NvdGFtdWNpZWJpZQo=
```

```
> echo czescradekcotamuciebie | base64 -w 6
Y3plc2
NyYWRl
a2NvdG
FtdWNp
ZWJpZQ
o=
```

I guess then it's ok to do it like you're doing it. Thanks for confirming!

sjchmiela (approved):

packages/steps/bin/set-env
radoslawkrzemien: I wonder what is the cause of that. Guess it's better to be safe than sorry at least for now

PR #493, opened 2025-01-14
[eas-build-job] Allow `steps` in build jobs

sjchmiela:
# Why

For workflows it'll be much easier to let users define `steps` in the workflow file instead of having to do the custom build config and stuff.

# How

Added ability to pass `steps` in build jobs. Alongside, we need to send `workflowInteprolationContext`, so added that too.

# Test Plan

Tested manually by running both manual and GitHub workflow build jobs. Also ran a custom build and a regular build from `eas-cli`.

### DIFF
--- a/packages/eas-build-job/src/generic.ts
+++ b/packages/eas-build-job/src/generic.ts
@@ -25,7 +25,8 @@ export namespace Generic {
     cocoapods: z.string().optional(),
   });
 
-  const CommonJobZ = z.object({
+  export type Job = z.infer<typeof JobZ>;
+  export const JobZ = z.object({
     projectArchive: ArchiveSourceSchemaZ,
     secrets: z.object({
       robotAccessToken: z.string(),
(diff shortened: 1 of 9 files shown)


packages/eas-build-job/src/generic.ts
```diff
@@ -42,25 +43,11 @@ export namespace Generic {
...
     steps: z.array(StepZ).min(1),
     outputs: z.record(z.string()).optional(),
   });
 
-  export type Job = z.infer<typeof JobZ>;
-  export const JobZ = z.union([PathJobZ, StepsJobZ]);
-
   export type PartialJob = z.infer<typeof PartialJobZ>;
-  export const PartialJobZ = z.union([PathJobZ.partial(), StepsJobZ.partial()]);
+  export const PartialJobZ = JobZ.partial();
```
szdziedzic, line 52: I don't know why but I believe doing so broke some types in Turtle and that's why I did this weird union with partials inside as a workaround 🤔

szdziedzic (approved):
I believe it looks reasonable

PR #495, opened 2025-01-16
[build-tools] resolve correct projectRootDirectory during eas build:internal

szdziedzic:
# Why

We resolve incorrect and override `projectRootDirectory` when running `eas build:internal` for GCS builds that use NoVcsClient in EAS CLI as VCS client. Adding `EAS_PROJECT_ROOT` env var should fix it.

# How

Use `EAS_PROJECT_ROOT` env var if `EAS_NO_VCS` is used

# Test Plan

Tests

### DIFF
--- a/packages/build-tools/src/common/easBuildInternal.ts
+++ b/packages/build-tools/src/common/easBuildInternal.ts
@@ -23,11 +23,13 @@ export async function runEasBuildInternalAsync<TJob extends BuildJob>({
   logger,
   env,
   cwd,
+  projectRootOverride,
 }: {
   job: TJob;
   logger: bunyan;
   env: BuildStepEnv;
   cwd: string;
+  projectRootOverride?: string;
 }): Promise<{
   newJob: TJob;
   newMetadata: Metadata;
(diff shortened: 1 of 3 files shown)


szdziedzic:
* **#495** <a href="https://app.graphite.dev/github/pr/expo/eas-build/495?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-build/495?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sjchmiela (changes requested):
Could we always pass `EAS_PROJECT_ROOT` (no matter `EAS_NO_VCS`)? I think we can expect `ctx.buildDirectory` / `ctx.projectTargetDirectory` to always be correct?

Also, why `buildDirectory` in one place and `projectTargetDirectory` in the other?

szdziedzic:
`buildDirectory` is a place where project archive is located in "standard" build process and `projectTargetDirectory` is the place where project archive lands after checkout is done for custom builds

szdziedzic:
>Could we always pass EAS_PROJECT_ROOT (no matter EAS_NO_VCS)? I think we can expect ctx.buildDirectory / ctx.projectTargetDirectory to always be correct?

I played it safely in case it has some unknown side effects :D

sjchmiela (approved):

PR #497, opened 2025-01-16
[steps] Add better workflow support for if condition

sjchmiela:
# Why

We want to support `if` conditions on individual steps in workflows.

# How

- adjusted expected template literals for `ifCondition`
- adjusted interpolation context used for `ifCondition`
- (_somewhat_ unrelated) added descriptions to `StepZ` Zod schema so we can replace `www` copies with using upstream

# Test Plan

Added some tests.

### DIFF
--- a/packages/steps/src/BuildStep.ts
+++ b/packages/steps/src/BuildStep.ts
@@ -308,17 +308,14 @@ export class BuildStep extends BuildStepOutputAccessor {
 
     let ifCondition = this.ifCondition;
 
-    if (ifCondition.startsWith('${') && ifCondition.endsWith('}')) {
+    if (ifCondition.startsWith('${{') && ifCondition.endsWith('}}')) {
+      ifCondition = ifCondition.slice(3, -2);
+    } else if (ifCondition.startsWith('${') && ifCondition.endsWith('}')) {
       ifCondition = ifCondition.slice(2, -1);
     }
 
     return Boolean(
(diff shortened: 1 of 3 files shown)


packages/steps/src/BuildStep.ts
```diff
@@ -308,17 +308,14 @@ export class BuildStep extends BuildStepOutputAccessor {
 
     let ifCondition = this.ifCondition;
 
-    if (ifCondition.startsWith('${') && ifCondition.endsWith('}')) {
+    if (ifCondition.startsWith('${{') && ifCondition.endsWith('}}')) {
```
szdziedzic, line 311: do we allow for `${{ sth }}` pattern in other places like inside of `run`? If we don't it's strange to support it only for `if`

packages/steps/src/BuildStep.ts
```diff
@@ -331,6 +328,7 @@ export class BuildStep extends BuildStepOutputAccessor {
           runtimePlatform: this.ctx.global.runtimePlatform,
           ...this.ctx.global.staticContext,
         },
+        ...this.getInterpolationContext(),
```
szdziedzic, line 331: this includes deleted things, right?

szdziedzic (approved):

packages/steps/src/BuildStep.ts
sjchmiela: Yea

PR #512, opened 2025-02-21
Add eas/download_build function

khamilowicz:
# Why

[ENG-14848: Add `eas/download_artifact`](https://linear.app/expo/issue/ENG-14848/add-easdownload-artifact)

User should be able to download build artifacts in a workflow or a custom build step.

# How

- Implemented `eas/download_build' function, which downloads a file and optionally extracts it. The path to the file (or a directory with the unpacked content) is returned:

# Test Plan

1. Link eas build to the local turtle code
2. start local `www` and `turtle-v2` instances
3. Submit the following custom build

```
build:
  name: Download and display
  steps:
      - eas/download_build:
           id: download_build
           inputs:
               build_id: build-id # replace with a build id
               fail_on_error: false
       - run: ls -la ${ steps.download_build.artifact_path }
```

# Deploy plan

1. https://github.com/expo/universe/pull/18902
2. https://github.com/expo/eas-build/pull/512

### DIFF
--- a/packages/build-tools/src/steps/functions/download_build.ts
+++ b/packages/build-tools/src/steps/functions/download_build.ts
@@ -1,0 +1,82 @@
+import {
+  BuildFunction,
+  BuildStepInput,
+  BuildStepInputValueTypeName,
+  BuildStepOutput,
+} from '@expo/steps';
+
+export function createDownloadBuildFunction(): BuildFunction {
+  return new BuildFunction({
+    namespace: 'eas',
+    id: 'download_build',
+    name: 'Download build',
+    inputProviders: [
+      BuildStepInput.createProvider({
+        id: 'build_id',
+        required: true,
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
+      BuildStepInput.createProvider({
+        id: 'fail_on_error',
+        required: false,
+        allowedValueTypeName: BuildStepInputValueTypeName.BOOLEAN,
+        defaultValue: true,
+      }),
+    ],
+    outputProviders: [
+      BuildStepOutput.createProvider({
+        id: 'artifact_path',
+        required: false,
+      }),
+    ],
+    command: `
+    BUILD_ID='\${ inputs.build_id }'
+    FAIL_ON_ERROR='\${ inputs.fail_on_error }'
+    APPLICATION_ARCHIVE_URL="\${__API_SERVER_URL%/}/v2/artifacts/eas/$BUILD_ID"
+
+    echo "Downloading build from $APPLICATION_ARCHIVE_URL"
+    if ! STORED_FILE_URL=$(curl --remote-name --retry 3 --retry-delay 20 --location --fail -w %{url_effective} --silent --show-error "$APPLICATION_ARCHIVE_URL"); then
+    if [ "$FAIL_ON_ERROR" = "true" ]; then
+        echo "File not available."
+        exit 1
+        fi
+      echo "File not available, skipping..."
+      exit 0
+    fi
+    echo "done."
+    URL_ARCHIVE_NAME=$(basename "$APPLICATION_ARCHIVE_URL")
+    ARCHIVE_NAME=$(basename "\${STORED_FILE_URL%%\\?*}")
+    mv $URL_ARCHIVE_NAME $ARCHIVE_NAME
+    echo "Downloaded application archive to $ARCHIVE_NAME"
+
+    if [[ "$ARCHIVE_NAME" == *.tar.gz || "$ARCHIVE_NAME" == *.tgz ]]; then
+        EXTRACTION_DIR=$(mktemp -d)
+        echo "Extracting application archive..."
+        tar -xf "$ARCHIVE_NAME" -C "$EXTRACTION_DIR"
+        echo "Extracted application archive to $EXTRACTION_DIR"
+    fi
+
+    if [ -d "$EXTRACTION_DIR" ]; then
+        mkdir -p "artifacts"
+
+        find "$EXTRACTION_DIR" -name "*.ipa" -exec mv {} "artifacts/" ';'
+        ARTIFACT_PATH=$(find . -name '*.ipa')
+
+        if [ -z "$ARTIFACT_PATH" ]; then
+            find "$EXTRACTION_DIR" -name "*.apk" -exec mv {} "artifacts/" ';'
+            ARTIFACT_PATH="$(find . -name '*.apk')"
+        fi
+
+        if [ -z "$ARTIFACT_PATH" ]; then
+            find "$EXTRACTION_DIR" -name "*.aab" -exec mv {} "artifacts/" ';'
+            ARTIFACT_PATH="$(find . -name '*.aab')"
+        fi
+    else
+        ARTIFACT_PATH=$ARCHIVE_NAME
+    fi
+    ARTIFACT_PATH=$(realpath "$ARTIFACT_PATH")
+    set-output artifact_path "$ARTIFACT_PATH"
+    echo "Set output artifact path to $ARTIFACT_PATH"
+    `,
+  });
+}
--- a/packages/build-tools/src/steps/easFunctions.ts
+++ b/packages/build-tools/src/steps/easFunctions.ts
@@ -28,6 +28,7 @@ import { calculateEASUpdateRuntimeVersionFunction } from './functions/calculateE
 import { createRepackBuildFunction } from './functions/repack';
 import { eagerBundleBuildFunction } from './functions/eagerBundle';
 import { createSubmissionEntityFunction } from './functions/createSubmissionEntity';
+import { createDownloadBuildFunction } from './functions/download_build';
 
 export function getEasFunctions(ctx: CustomBuildContext): BuildFunction[] {
   const functions = [
@@ -38,6 +39,7 @@ export function getEasFunctions(ctx: CustomBuildContext): BuildFunction[] {
     createPrebuildBuildFunction(),
 
     configureEASUpdateIfInstalledFunction(),
+    createDownloadBuildFunction(),
     injectAndroidCredentialsFunction(),
     configureAndroidVersionFunction(),
     eagerBundleBuildFunction(),


packages/build-tools/src/steps/easFunctionGroups.ts
```diff
@@ -4,12 +4,15 @@ import { CustomBuildContext } from '../customBuildContext';
...
 import { createEasMaestroTestFunctionGroup } from './functionGroups/maestroTest';
+import { createEasSubmitBuildFunctionGroup } from './functionGroups/submit';
 
 export function getEasFunctionGroups(ctx: CustomBuildContext): BuildFunctionGroup[] {
   const functionGroups = [createEasMaestroTestFunctionGroup(ctx)];
 
   if (ctx.hasBuildJob()) {
-    functionGroups.push(...[createEasBuildBuildFunctionGroup(ctx)]);
+    functionGroups.push(
+      ...[createEasBuildBuildFunctionGroup(ctx), createEasSubmitBuildFunctionGroup(ctx)]
```
sjchmiela, line 14: We should make it so `eas/submit` does not require a build job to run. It currently runs in a generic job and it should remain like this.

When this lands we'll replace implementations in `WorkflowSubmitJob` with calls to `eas/submit`.

packages/build-tools/src/steps/functions/download_artifact.ts
```diff
@@ -0,0 +1,65 @@
...
+} from '@expo/steps';
+
+export function createDownloadArtifactFunction(): BuildFunction {
+  return new BuildFunction({
+    namespace: 'eas',
+    id: 'download_artifact',
+    name: 'Download artifact',
+    inputProviders: [
+      BuildStepInput.createProvider({
+        id: 'application_archive_url',
```
sjchmiela, line 15: I think it would be more interesting to use a `build_id` here. `eas/download_artifact` needs to allow us to get rid of `getShortApplicationArchiveUrl()` ([Linear task](https://linear.app/expo/issue/ENG-14849/replace-use-of-turtlebuildutilsgetshortapplicationarchiveurl-with)).

Actually, maybe this should be called `eas/download_build` instead?

This may require adding a new endpoint to some `www` router.

packages/build-tools/src/steps/functions/download_artifact.ts
```diff
@@ -0,0 +1,65 @@
...
+    APPLICATION_ARCHIVE_URL='\${ inputs.application_archive_url }'
+    if [ -z "$APPLICATION_ARCHIVE_URL" ]; then
+        echo "No application archive URL provided"
+        exit 0
+    fi
+    curl --remote-name --location --fail --silent --show-error "$APPLICATION_ARCHIVE_URL"
+    ARCHIVE_NAME=$(basename "$APPLICATION_ARCHIVE_URL")
+    echo "Downloaded application archive to $ARCHIVE_NAME"
+
+    EXTRACTOIN_DIR=$(mktemp -d)
```
sjchmiela, line 36: Did you test a `.tar.gz` scenario? Can we add integration tests for this? (There's a typo.)

packages/build-tools/src/steps/functions/download_artifact.ts
```diff
@@ -0,0 +1,65 @@
...
+        required: false,
+      }),
+    ],
+    command: `
+    APPLICATION_ARCHIVE_URL='\${ inputs.application_archive_url }'
+    if [ -z "$APPLICATION_ARCHIVE_URL" ]; then
+        echo "No application archive URL provided"
+        exit 0
+    fi
+    curl --remote-name --location --fail --silent --show-error "$APPLICATION_ARCHIVE_URL"
```
sjchmiela, line 32: One thing I was thinking of doing is making the `echo` nicer, like
```sh
echo -n "Downloading build from $URL..."
curl
echo " done."
```

packages/build-tools/src/steps/functions/download_artifact.ts
```diff
@@ -0,0 +1,65 @@
...
+        if [ -z "$ARTIFACT_PATH" ]; then
+            find "$EXTRACTION_DIR" -name "*.aab" -exec mv {} "artifacts/" ';'
+            ARTIFACT_PATH="$(find . -name '*.aab')"
+        fi
+
+        set-output artifact_path "$ARTIFACT_PATH"
+        echo "Set output artifact path to $ARTIFACT_PATH"
+    else
+        set-output artifact_path "$ARCHIVE_NAME"
+        echo "Set output artifact path to $ARCHIVE_NAME"
```
sjchmiela, line 61: Let's make these absolute.

packages/build-tools/src/steps/functions/download_artifact.ts
```diff
@@ -0,0 +1,65 @@
...
+    EXTRACTOIN_DIR=$(mktemp -d)
+
+    if [[ "$ARCHIVE_NAME" == *.tar.gz || "$ARCHIVE_NAME" == *.tgz ]]; then
+        tar -xf "$ARCHIVE_NAME" -C "$EXTRACTION_DIR"
+        echo "Extracted application archive to $EXTRACTION_DIR"
+    fi
+
+    if [ -d "$EXTRACTION_DIR" ]; then
+        find "$EXTRACTION_DIR" -name "*.ipa" -exec mv {} "artifacts/" ';'
+        ARTIFACT_PATH=$(find . -name '*.ipa')
```
sjchmiela, line 45: If we're making it reusable, let's support `*.app`s too. Note these are directories.

packages/build-tools/src/steps/functionGroups/submit.ts
```diff
@@ -0,0 +1,732 @@
...
+      BuildStepInput.createProvider({
+        id: 'is_verbose_fastlane_enabled',
+        required: false,
+        allowedValueTypeName: BuildStepInputValueTypeName.BOOLEAN,
+      }),
+    ],
+    createBuildStepsFromFunctionGroupCall: (globalCtx, { inputs }) => {
+      validateInputs(inputs);
+
+      const buildId = String(inputs.build_id.value ?? globalCtx.env.EAS_BUILD_ID);
```
sjchmiela, line 60: That's smart!!

packages/build-tools/src/steps/functionGroups/submit.ts
```diff
@@ -0,0 +1,732 @@
...
+      }),
+      BuildStepInput.createProvider({
+        id: 'application_archive_url',
+        required: false,
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
+      BuildStepInput.createProvider({
+        id: 'build_id',
+        required: false,
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
```
sjchmiela, line 44: We can't make it `defaultValue: ${ env.BUILD_ID }`, can we?

packages/build-tools/src/steps/functionGroups/submit.ts
```diff
@@ -0,0 +1,732 @@
...
+      } else {
+        return createStepsForAndroidSubmit(commandOptions, globalCtx);
+      }
+    },
+  });
+}
+
+function validateInputs(inputs: BuildStepInputById): void {
+  const providedArgs = [
+    inputs.build_id.value,
```
sjchmiela, line 93: Should we make sure this errors if none is provided and there's also no `env.BUILD_ID`?

sjchmiela (changes requested):

packages/build-tools/src/steps/functionGroups/submit.ts
```diff
@@ -0,0 +1,732 @@
...
+      env: { HOMEBREW_NO_AUTO_UPDATE: '1' },
+      fn: async (stepCtx, { env }) => {
+        await spawnAsync('/opt/homebrew/bin/brew', ['install', 'jq'], {
+          env,
+          logger: stepCtx.logger,
+          stdio: ['ignore', 'pipe', 'pipe'],
+        });
+      },
+    }),
+    new BuildStep(globalCtx, {
```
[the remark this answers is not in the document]
khamilowicz: Yes and no. Yes, thanks to this step we can avoid downloading the archive, but no, we can't skip this step. When we run `eas/build` and `eas/submit` in the same job, right after the other, `www` may not be synchronised yet with turtle, so the archive url is not visible yet. We can probably make this step a fallback ('if archiveUrl is not available, search local paths`), but IMO this is the expected behaviour.

packages/build-tools/src/steps/functionGroups/submit.ts
khamilowicz: no, we can't

packages/build-tools/src/steps/functionGroups/submit.ts
khamilowicz: 😊

packages/build-tools/src/steps/functionGroups/submit.ts
```diff
@@ -0,0 +1,732 @@
...
+        allowedValueTypeName: BuildStepInputValueTypeName.BOOLEAN,
+      }),
+    ],
+    createBuildStepsFromFunctionGroupCall: (globalCtx, { inputs }) => {
+      validateInputs(inputs);
+
+      const buildId = String(inputs.build_id.value ?? globalCtx.env.EAS_BUILD_ID);
+      const profile = inputs.profile.value ? String(inputs.profile.value) : undefined;
+      const isVerboseFastlaneEnabled = inputs.is_verbose_fastlane_enabled.value
+        ? String(inputs.is_verbose_fastlane_enabled.value)
```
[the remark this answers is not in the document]
khamilowicz: `BuildStep.env` accepts only strings (we use it only there), and we validate its booleanness in BuildStepInputProvider. I don't think we need to cast it to a string closer to the use place.

packages/build-tools/src/steps/functions/download_build.ts
```diff
@@ -0,0 +1,71 @@
...
+      }),
+    ],
+    command: `
+    BUILD_ID='\${ inputs.build_id }'
+    APPLICATION_ARCHIVE_URL="\${__API_SERVER_URL}v2/artifacts/eas/$BUILD_ID"
+
+    echo "Downloading build from $APPLICATION_ARCHIVE_URL"
+    if ! STORED_FILE_URL=$(curl --remote-name --location --fail -w %{url_effective} --silent --show-error "$APPLICATION_ARCHIVE_URL"); then
+      echo "File not available, skipping..."
+      exit 0
```
sjchmiela, line 33: I think we should throw? `eas/download_build` is more serious than `eas/download_cache` or sth. Cache is expected to be rebuildable, build is a build.

packages/build-tools/src/steps/functions/download_build.ts
```diff
@@ -0,0 +1,71 @@
...
+            find "$EXTRACTION_DIR" -name "*.apk" -exec mv {} "artifacts/" ';'
+            ARTIFACT_PATH="$(find . -name '*.apk')"
+        fi
+
+        if [ -z "$ARTIFACT_PATH" ]; then
+            find "$EXTRACTION_DIR" -name "*.aab" -exec mv {} "artifacts/" ';'
+            ARTIFACT_PATH="$(find . -name '*.aab')"
+        fi
+    else
+        ARTIFACT_PATH=$ARCHIVE_NAME
```
sjchmiela, line 64: What happens if the build is a simulator build?

packages/build-tools/src/steps/functions/download_build.ts
```diff
@@ -0,0 +1,71 @@
...
+        id: 'artifact_path',
+        required: false,
+      }),
+    ],
+    command: `
+    BUILD_ID='\${ inputs.build_id }'
+    APPLICATION_ARCHIVE_URL="\${__API_SERVER_URL}v2/artifacts/eas/$BUILD_ID"
+
+    echo "Downloading build from $APPLICATION_ARCHIVE_URL"
+    if ! STORED_FILE_URL=$(curl --remote-name --location --fail -w %{url_effective} --silent --show-error "$APPLICATION_ARCHIVE_URL"); then
```
sjchmiela, line 31: Is there a reason you go through `-w url_effective` and current submit job "just works"? What's the benefit?

packages/build-tools/src/steps/functionGroups/submit.ts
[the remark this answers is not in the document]
sjchmiela: Hmm, you're right. However, as we move off of Turtle and on to `workflow-orchestration`, we'll be handling uploads differently. Basically, in workflows, the application archive is saved on the Turtle Build entity as soon as it is uploaded.

https://github.com/expo/turtle-v2/blob/cb111ccb16a39ab34d6d275678b541601a40ca15/src/services/worker/src/upload.ts#L51-L53

---

Let's maybe not worry about `eas/build` + `eas/submit` in the same job for now and remove this step altogether. There's [a clear use case](https://bsky.app/profile/yef.im/post/3lj7mt525e22x) for `eas/submit` in its own submit job (let the user build the necessary packages and then call `eas/submit`). Let's land this and then start improving. This should simplify things a bit and make easier to review (being closer to the originals).

packages/build-tools/src/steps/functionGroups/submit.ts
```diff
@@ -0,0 +1,601 @@
...
+        allowedValueTypeName: BuildStepInputValueTypeName.BOOLEAN,
+      }),
+    ],
+    createBuildStepsFromFunctionGroupCall: (globalCtx, { inputs }) => {
+      validateInputs(inputs, globalCtx.env);
+
+      const buildId = String(inputs.build_id.value ?? globalCtx.env.EAS_BUILD_ID);
+      const profile = String(inputs.profile.value ?? globalCtx.env.EAS_BUILD_PROFILE);
+
+      const isVerboseFastlaneEnabled = isUndefined(inputs.is_verbose_fastlane_enabled.value)
```
sjchmiela, line 59:
```suggestion
      const isVerboseFastlaneEnabled = inputs.is_verbose_fastlane_enabled.value == undefined
```

packages/build-tools/src/steps/functionGroups/submit.ts
```diff
@@ -0,0 +1,601 @@
...
+      outputs: [
+        new BuildStepOutput(globalCtx, {
+          stepDisplayName: 'Find existing application artifacts',
+          id: 'archive_path',
+          required: false,
+        }),
+      ],
+      fn: async (stepCtx, { outputs }) => {
+        if (userApplicationArchivePath) {
+          fs.accessSync(userApplicationArchivePath);
```
sjchmiela, line 139: We tend to use `fs/promises` over synchronous versions of functions.

sjchmiela (changes requested):

packages/build-tools/src/steps/functions/download_build.ts
```diff
@@ -0,0 +1,71 @@
...
+    ],
+    outputProviders: [
+      BuildStepOutput.createProvider({
+        id: 'artifact_path',
+        required: false,
+      }),
+    ],
+    command: `
+    BUILD_ID='\${ inputs.build_id }'
+    APPLICATION_ARCHIVE_URL="\${__API_SERVER_URL}v2/artifacts/eas/$BUILD_ID"
```
szdziedzic, line 28: we assume that `__API_SERVER_URL` will always have `/` at the end

isn't it risky? can we do something like `path.join`?

packages/build-tools/src/steps/functions/download_build.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    echo "Downloading build from $APPLICATION_ARCHIVE_URL"
+    if ! STORED_FILE_URL=$(curl --remote-name --location --fail -w %{url_effective} --silent --show-error "$APPLICATION_ARCHIVE_URL"); then
+      echo "File not available, skipping..."
+      exit 0
+    fi
```
szdziedzic, line 34: if we fail to download the file inside of the `eas/download_file` function shouldn't we fail? it seems to be the core thing this step does

proceeding with successful `eas/download_build` and having some workflows fail in later steps due to missing build afterward seems like a hard thing to debug

as a user, I would prefer it to fail straightaway

packages/build-tools/src/steps/functions/download_build.ts
```diff
@@ -0,0 +1,71 @@
...
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
+    ],
+    outputProviders: [
+      BuildStepOutput.createProvider({
+        id: 'artifact_path',
+        required: false,
+      }),
+    ],
+    command: `
```
szdziedzic, line 26: It doesn't need to be done for v1, let's iterate on it later on, but I would personally suggest having `eas/download_build` as TS function vs bash script

I believe using TS, libraries and our other existing utils would make it less error-prone

packages/build-tools/src/steps/functionGroups/submit.ts
```diff
@@ -0,0 +1,601 @@
...
+        try {
+          const applicationArchives = await findArtifacts({
+            rootDir: stepCtx.workingDirectory,
+            patternOrPath: 'android/app/build/outputs/**/*.{apk,aab}',
+            logger: stepCtx.logger,
+          });
+
+          if (applicationArchives.length === 0) {
+            return;
+          }
```
szdziedzic, line 393: wouldn't this execution path result in exiting without printing any logs?

packages/build-tools/src/steps/functionGroups/submit.ts
```diff
@@ -0,0 +1,601 @@
...
+            throw new Error(
+              'Found more than one application archive. Provide `application_archive_path` input.'
+            );
+          }
+          outputs.archive_path.set(applicationArchives[0]);
+        } catch (err: any) {
+          stepCtx.logger.info({ err }, `Failed to find existing application artifacts.`);
+        }
+      },
+    }),
```
szdziedzic, line 409: I believe we can move it to a function we can reuse for both Android and iOS or just generally `eas/` custom function

szdziedzic (changes requested):
We can do it in the next iteration so as not to hold the merging of this PR, but I would prefer us to use more TS vs bash scripting. I mean it seems like it's easier to miss/break something in these long 200 sections of bash scripts vs when using well-structured TS code divided into functions and so on.

packages/build-tools/src/steps/functions/download_build.ts
khamilowicz: Yes, you are right. Now that we don't rely on `synchronizer` to update `turtle_build` on `www` on time, this should always return a build artifact.

packages/build-tools/src/steps/functions/download_build.ts
khamilowicz: I use url_effective to obtain the ultimate url after redirects, which would point to a bucket, where the key would be the uploaded file name, which should contain the file extension. Currently TurtleBuildUtils do it the same way (although based on the stored data instead of the actual bucket key) https://github.com/expo/universe/blob/8fb9114f3b1b3b1ee38c3c637ce7f62664de7354/server/www/src/data/entities/turtlebuild/TurtleBuildUtils.ts#L989

sjchmiela (changes requested):
https://github.com/expo/eas-build/pull/512#pullrequestreview-2709777214

khamilowicz:
* **#524** <a href="https://app.graphite.dev/github/pr/expo/eas-build/524?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#512** <a href="https://app.graphite.dev/github/pr/expo/eas-build/512?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-build/512?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #515, opened 2025-03-16
[build-tools] Use `rawValue` when collecting step outputs

sjchmiela:
# Why

When testing another pull request and running failing workflow runs on my machine I noticed uploading outputs may fail if the workflow hasn't set outputs for the skipped steps.

Zrzut ekranu 2025-03-16 o 15 44 10

# How

Instead of using `value` which is erroring if the output is required, but not set, use `rawValue` when collecting outputs. The only difference between them is that validation.

https://github.com/expo/eas-build/blob/ece9c10507a155f9b628c4ba4cbfb82780e3f599/packages/steps/src/BuildStepOutput.ts#L52-L63

# Test Plan

Later workflow run uploaded outputs successfully.

### DIFF
--- a/packages/build-tools/src/utils/outputs.ts
+++ b/packages/build-tools/src/utils/outputs.ts
@@ -76,7 +76,7 @@ export function getStepOutputsAsObject(step: BuildStep): {
   outputs: Record<string, string | undefined>;
 } {
   const outputs = Object.fromEntries(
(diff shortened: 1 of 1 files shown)


szdziedzic (commented):
would it make sense to change `value` logic to skip throwing error if step is skipped?

it seems like it safeguards us from not setting required outputs, which can be good?

sjchmiela:
https://github.com/expo/eas-build/pull/518 fixes this and other things

PR #516, opened 2025-03-18
[eas-build-job] [ENG-15011] Add label to workflow interpolation context

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-15011/proper-ui-for-github-pr-labeled-workflow-runs
To allow for using labels in conditional expressions
Companion to: https://github.com/expo/universe/pull/18734 and https://github.com/expo/universe/pull/19216

# How

Added an optional `github.labell` field to the workflow interpolation context

# Test Plan

### DIFF
--- a/packages/eas-build-job/src/common.ts
+++ b/packages/eas-build-job/src/common.ts
@@ -189,6 +189,7 @@ export const StaticWorkflowInterpolationContextZ = z.object({
       ref: z.string(),
       ref_name: z.string(),
       ref_type: z.string(),
+      label: z.string().optional(),
     })
     // We need to .optional() to support jobs that are not triggered by a GitHub event.
     .optional(),


packages/eas-build-job/src/common.ts
```diff
@@ -189,6 +189,7 @@ export const StaticWorkflowInterpolationContextZ = z.object({
       ref: z.string(),
       ref_name: z.string(),
       ref_type: z.string(),
+      label: z.string().optional(),
```
sjchmiela, line 192: Is this what GitHub exposes in its event payload? Like, do people do `if: ${{ github.label == '...' }}` in their GHA workflows?

Ideally we would at least match what GitHub is doing so users migrating from it don't need to change code?

radoslawkrzemien: I'll check

radoslawkrzemien: GHA uses `github.event.label.name`. A bit complex compared to simple `github.[prop]` but would be consistent with GHA

szdziedzic (approved):

packages/eas-build-job/src/common.ts
radoslawkrzemien: @sjchmiela which option you think is better?

sjchmiela: Let's add both? Wdyt?

radoslawkrzemien: Cool with me

sjchmiela (approved):

sjchmiela (approved):

PR #518, opened 2025-03-20
[build-tools] Improve interpolation context on `job.outputs`

sjchmiela:
# Why

I used `fromJSON` in `outputs` in https://github.com/expo/universe/pull/19328, but had to revert it in https://github.com/expo/universe/pull/19361, because `uploadJobOutputsToWwwAsync` was using custom interpolation context when generating values.

# How

Instead of using custom `{ steps }`, used global context to get static context and `JobInterpolationContext` type to ensure we provide `fromJSON`, `toJSON` and all.

# Test Plan

Ran a workflow 

```yaml
jobs:
  publish_update:
    outputs:
      updates_json: ${{ steps.step.outputs.updates_json }}
      first_update_group_id: ${{ fromJSON(steps.step.outputs.updates_json || '[]')[0].group }}
    steps:
      - uses: eas/checkout
      - uses: eas/install_node_modules
      - id: step
        run: |
          UPDATES_JSON=$(npx -y eas-cli@latest update --non-interactive --json --auto --branch test)
          set-output updates_json "$UPDATES_JSON"
```

and got

Zrzut ekranu 2025-03-20 o 18 21 47

### DIFF
--- a/packages/build-tools/src/generic.ts
+++ b/packages/build-tools/src/generic.ts
@@ -42,8 +42,7 @@ export async function runGenericJobAsync(
   const runResult = await asyncResult(workflow.executeAsync());
 
   await ctx.runBuildPhase(BuildPhase.COMPLETE_JOB, async () => {
-    await uploadJobOutputsToWwwAsync(ctx, {
-      steps: workflow.buildSteps,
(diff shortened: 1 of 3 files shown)


szdziedzic (approved):
lgtm

PR #525, opened 2025-04-04
[build-tools] [ENG-15188] Cancel hanging workflows

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-15188/replicate-the-cancel-hanging-build-functionality-for-workflows

# How

Added warn and kill timeouts for workflow execution, which get reset after logs are printed, and removed after workflow step finishes. If they time out, first a warning is printed and then process gets killed. Similar solution works for builds

# Test Plan

Automated tests
TODO: Manual tests


packages/steps/src/BuildStep.ts
```diff
@@ -360,6 +370,27 @@ export class BuildStep extends BuildStepOutputAccessor {
       toJSON: (value: unknown) => JSON.stringify(value),
     };
   }
+
+  private setCommandSpawnTimeouts(spawnPromise: SpawnPromise<SpawnResult>): void {
+    this.commandWarnTimeout = setTimeout(() => {
+      this.ctx.logger.warn(
+        'Command takes longer then expected and it did not produce any logs in the past 15 minutes. Consider evaluating your command for possible issues'
```
sjchmiela, line 377:
```suggestion
        'Command takes longer then expected and it did not produce any logs in the past 15 minutes. Consider evaluating your command for possible issues.'
```

packages/steps/src/BuildStep.ts
```diff
@@ -360,6 +370,27 @@ export class BuildStep extends BuildStepOutputAccessor {
...
+      );
+    }, BUILD_STEP_WARN_TIMEOUT_MS);
+
+    this.commandKillTimeout = setTimeout(async () => {
+      this.commandTimedOut = true;
+      this.ctx.logger.error(
+        'Command takes a very long time and it did not produce any logs in the past 30 minutes. Most likely an unexpected error happened which caused the process to hang and it will be terminated'
```
sjchmiela, line 384:
```suggestion
        'Command takes a very long time and it did not produce any logs in the past 30 minutes. Most likely an unexpected error happened which caused the process to hang and it will be terminated.'
```

sjchmiela (changes requested):
Can you add some tests too?

packages/steps/src/BuildStep.ts
```diff
@@ -381,13 +412,41 @@ export class BuildStep extends BuildStepOutputAccessor {
...
       `Executing script: ${shellCommand}${args !== undefined ? ` ${args.join(' ')}` : ''}`
     );
-    await spawnAsync(shellCommand, args ?? [], {
-      cwd: this.ctx.workingDirectory,
-      logger: this.ctx.logger,
-      env: this.getScriptEnv(),
-      // stdin is /dev/null, std{out,err} are piped into logger.
-      stdio: ['ignore', 'pipe', 'pipe'],
-    });
+    this.commandTimedOut = false;
```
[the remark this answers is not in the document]
radoslawkrzemien: Just readability, if all is defined within the function executing the command over half of the function body is taken by timeout handing, which is not the main purpose of the function. Having those as properties allows for extracting the timeout setup to a separate private method

packages/steps/src/BuildStep.ts
```diff
@@ -360,6 +370,27 @@ export class BuildStep extends BuildStepOutputAccessor {
...
+
+    this.commandKillTimeout = setTimeout(async () => {
+      this.commandTimedOut = true;
+      this.ctx.logger.error(
+        'Command takes a very long time and it did not produce any logs in the past 30 minutes. Most likely an unexpected error happened which caused the process to hang and it will be terminated'
+      );
+      const ppid = nullthrows(spawnPromise.child.pid);
+      const pids = await getParentAndDescendantProcessPidsAsync(ppid);
+      pids.forEach((pid) => {
+        process.kill(pid);
```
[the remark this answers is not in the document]
radoslawkrzemien: I *think* it should be, but we do it the same way with `Install Dependencies` and `Install Pods` build steps and I seem to vaguely remember there was a reason for it, but I can't recall it now

packages/steps/src/BuildStep.ts
```diff
@@ -57,6 +60,10 @@ export type BuildStepFunction = (
 const UUID_REGEX =
   /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/;
 
+const BUILD_STEP_WARN_TIMEOUT_MS = 15 * 60 * 1000;
+const BUILD_STEP_KILL_TIMEOUT_MS = 30 * 60 * 1000;
```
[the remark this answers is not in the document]
radoslawkrzemien: I'll look into it

packages/steps/src/BuildStep.ts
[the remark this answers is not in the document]
sjchmiela: Maybe we should move this (setting up timeouts and catching-and-throwing-speciifc-error) to `spawnAsync`?

packages/steps/src/BuildStep.ts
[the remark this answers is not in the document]
radoslawkrzemien: GHA has `jobs.<job_id>.timeout-minutes` and `jobs.<job_id>.steps[*].timeout-minutes`, although it's not the same as what we are doing since it only looks at total job/step execution time - even if the job/step is alive and printing, if it passes the timeout it gets cancelled. So in this case I think we shouldn't try to mimick GHA as it would be confusing to people familiar with it, maybe something like `no-logs-timeout`? WDYT?

packages/steps/src/BuildStep.ts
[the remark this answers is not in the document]
radoslawkrzemien: That might work. Are you aware of any cases where we use `spawnAsync` and might **not** want to have timeouts?

packages/steps/src/utils/shell/spawn.ts
```diff
@@ -1,45 +1,161 @@
...
+type NoLogsTimeoutOptions = {
+  warn?: {
+    timeoutMinutes?: number;
+    message?: string;
+  };
+  kill?: {
+    timeoutMinutes?: number;
+    message?: string;
+    errorClass?: IErrorClass;
+    errorMessage?: string;
```
sjchmiela, line 26: Is there a reason we need to customize `errorClass` and `errorMessage`? Couldn't we just pass in `warnTimeoutMs` and `killTimeoutMs`?

packages/steps/src/utils/shell/spawn.ts
```diff
@@ -1,45 +1,161 @@
...
+  ): void {
+    if (noLogsTimeout.warn) {
+      const warnTimeoutMinutes =
+        noLogsTimeout.warn.timeoutMinutes ?? SPAWN_WARN_TIMEOUT_DEFAULT_MINUTES;
+      spawnWarnTimeout = setTimeout(
+        () => {
+          logger.warn(
+            noLogsTimeout.warn?.message ??
+              SPAWN_WARN_TIMEOUT_DEFAULT_MESSAGE.replace(
+                '${minutes}',
```
sjchmiela, line 86: Let's make this simpler.

packages/steps/src/utils/shell/spawn.ts
```diff
@@ -1,45 +1,161 @@
...
+    }
+    spawnResult = await spawnPromise;
+  } catch (err: any) {
+    if (spawnTimedOut) {
+      const spawnKillTimeout =
+        noLogsTimeout?.kill?.timeoutMinutes ?? SPAWN_KILL_TIMEOUT_DEFAULT_MINUTES;
+      const errorMessage =
+        noLogsTimeout?.kill?.errorMessage ??
+        SPAWN_KILL_TIMEOUT_DEFAULT_ERROR_MESSAGE.replace('${minutes}', spawnKillTimeout.toString());
+      const ErrorClass = noLogsTimeout?.kill?.errorClass ?? Error;
```
sjchmiela, line 148: Can we use some well-known error class for this? Maybe we want to use `UserFacingError` for this, or add a new subclass of `UserFacingError` in `eas-build-job` and use that?

packages/steps/src/utils/shell/spawn.ts
```diff
@@ -1,45 +1,161 @@
...
+            }
+          },
+        };
+        pipeSpawnOutput(logger, spawnPromise.child, optionsWithCallback);
+        setCommandSpawnTimeouts(noLogsTimeout, logger, spawnPromise);
+      } else {
+        pipeSpawnOutput(logger, spawnPromise.child, options);
+      }
+    }
+    spawnResult = await spawnPromise;
```
sjchmiela, line 140: Hmm, I think in some places we don't `await` the `SpawnPromise` by design. E.g. [here](https://github.com/expo/eas-build/blob/e901e53419a181e4d1bdfcc3bd8ac4a952f0bbac/packages/build-tools/src/steps/functions/startAndroidEmulator.ts#L67-L80).

Instead of `try-catch` and `await`ing the `spawnPromise`, could we attach the `catch` as `spawnPromise.catch(err => ...)`?

sjchmiela (changes requested):

packages/steps/src/utils/shell/spawn.ts
radoslawkrzemien: We don't *need* to, if not passed a default error with a default message will be thrown. But I thought it might be useful to customise them depending on where the functionality is used (workflow command, build step, etc.) and have the appropriate message

PR #527, opened 2025-04-24
[build-tools][ENG-13286] Use EAS_BUILD_ENABLE_VERBOSE_LOGGING variable to enable more verbose logging.

khamilowicz:
# Why

[ENG-13286: Provide an option for EAS Build that indicates we should use the verbose logging version of the install command](https://linear.app/expo/issue/ENG-13286/provide-an-option-for-eas-build-that-indicates-we-should-use-the)

User should be able to easily enable verbose logs for build phase for easier debugging.

# How

- added verbose flag to `installDependencies` (js package manager)
- added verbose flag to pod install commands
- added `--info` to gradlew

# Test Plan

See https://github.com/expo/eas-cli/pull/3000

# Deploy

1. https://github.com/expo/eas-build/pull/527
2. https://github.com/expo/eas-cli/pull/3000

### DIFF
--- a/packages/build-tools/src/common/installDependencies.ts
+++ b/packages/build-tools/src/common/installDependencies.ts
@@ -20,6 +20,9 @@ export async function installDependenciesAsync<TJob extends Job>(
       args = ['install', '--no-immutable', '--inline-builds'];
     }
   }
+  if (ctx.env['EAS_VERBOSE'] === '1') {
+    args = [...args, '--verbose'];
+  }
   logger?.info(`Running "${ctx.packageManager} ${args.join(' ')}" in ${cwd} directory`);
   return {
     spawnPromise: spawn(ctx.packageManager, args, {
--- a/packages/build-tools/src/android/gradle.ts
+++ b/packages/build-tools/src/android/gradle.ts
@@ -31,7 +31,9 @@ export async function runGradleCommand(
 ): Promise<void> {
   logger.info(`Running 'gradlew ${gradleCommand}' in ${androidDir}`);
   await fs.chmod(path.join(androidDir, 'gradlew'), 0o755);
-  const spawnPromise = spawn('bash', ['-c', `./gradlew ${gradleCommand}`], {
+  const verboseFlag = ctx.env['EAS_VERBOSE'] === '1' ? '--info' : '';
+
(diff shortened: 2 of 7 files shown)


packages/build-tools/src/common/installDependencies.ts
```diff
@@ -20,6 +20,9 @@ export async function installDependenciesAsync<TJob extends Job>(
       args = ['install', '--no-immutable', '--inline-builds'];
     }
   }
+  if (ctx.env['EAS_BUILD_ENABLE_VERBOSE_LOGGING'] === '1') {
+    args = [...args, '--verbose'];
```
sjchmiela, line 24: Did you confirm that all package managers support `--verbose` flag? (It could be that some package manager does not and throws an error if unknown flag is provided.)

sjchmiela (changes requested):
I wonder why choose `EAS_BUILD_ENABLE_VERBOSE_LOGGING`? Could we use `DEBUG`? `VERBOSE`? Maybe something else, but shorter (thinking of [this](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/troubleshooting-workflows/enabling-debug-logging))?

packages/build-tools/src/common/installDependencies.ts
khamilowicz: Good catch. All supported package managers support this flag.

khamilowicz:
@sjchmiela I chose a non-generic name intentionally to avoid potential conflicts with existing environment variable–based logging tools. By using a more specific and verbose name, we can ensure that logging is only enabled for the commands we explicitly want, without needing to worry about cases like `DEBUG=true jq`, which could unintentionally trigger unrelated logging behavior.

sjchmiela (approved):
> I chose a non-generic name intentionally to avoid potential conflicts with existing environment variable–based logging tools. By using a more specific and verbose name, we can ensure that logging is only enabled for the commands we explicitly want, without needing to worry about cases like DEBUG=true jq, which could unintentionally trigger unrelated logging behavior.

Makes sense. However, I wish we didn't include `_BUILD_` anymore since this can run in workflows too. How about `EAS_VERBOSE=1`?

sjchmiela (approved):

PR #530, opened 2025-04-28
[build-tools] Use "`--frozen-lockfile`" for SDK53+

sjchmiela:
Takeover of https://github.com/expo/eas-build/pull/463

# Why

https://exponent-internal.slack.com/archives/C02123T524U/p1731420965266849?thread_ts=1731420501.834709&cid=C02123T524U
https://exponent-internal.slack.com/archives/C02123T524U/p1731421835879399

We want to move over to using "frozen lockfile" when installing dependencies on EAS.

# How

Use "frozen lockfile" if :
- `EAS_NO_FROZEN_LOCKFILE` environment variable is _not_ present AND
- we know the SDK version is not <= 52 AND
- we know the React Native version is not <= 0.79.

Don't use frozen lockfile after prebuild as prebuild can sometimes modify `the package.json`.

This is technically a breaking change? Which is why I'm making it for SDK53+ (not SDK52+). Can't do anything about the non-SDK projects though… we can do the switch now or never…

# Test Plan

- [x] tested locally

Left: SDK53, right: SDK52

Zrzut ekranu 2025-04-28 o 13 25 17

### DIFF
--- a/packages/build-tools/src/common/prebuild.ts
+++ b/packages/build-tools/src/common/prebuild.ts
@@ -32,10 +32,13 @@ export async function prebuildAsync<TJob extends BuildJob>(
     packageManager: ctx.packageManager,
   });
   const installDependenciesSpawnPromise = (
-    await installDependenciesAsync(ctx, {
+    await installDependenciesAsync({
+      packageManager: ctx.packageManager,
+      env: ctx.env,
       logger,
       cwd: resolvePackagerDir(ctx),
-      withoutFrozenLockfile: true, // prebuild should can modify package.json in some cases
+      // prebuild sometimes modifies package.json, so we don't want to use frozen lockfile
+      useFrozenLockfile: false,
     })
   ).spawnPromise;
   await installDependenciesSpawnPromise;
--- a/packages/build-tools/src/common/prebuild.ts
+++ b/packages/build-tools/src/common/prebuild.ts
@@ -35,6 +35,7 @@ export async function prebuildAsync<TJob extends BuildJob>(
     await installDependenciesAsync(ctx, {
       logger,
       cwd: resolvePackagerDir(ctx),
+      withoutFrozenLockfile: true, // prebuild should can modify package.json in some cases
     })
   ).spawnPromise;
   await installDependenciesSpawnPromise;
--- a/packages/build-tools/src/steps/functions/installNodeModules.ts
+++ b/packages/build-tools/src/steps/functions/installNodeModules.ts
@@ -9,7 +9,7 @@ import {
   PackageManager,
   resolvePackageManager,
 } from '../../utils/packageManager';
-import { isUsingYarn2 } from '../../utils/project';
+import { isUsingModernYarnVersion } from '../../utils/project';
 
 export function createInstallNodeModulesBuildFunction(): BuildFunction {
   return new BuildFunction({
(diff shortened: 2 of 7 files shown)


Kudo (approved):
thanks for following up this feature 💛

kadikraman (approved):
Looks good to me!

packages/build-tools/src/steps/functions/installNodeModules.ts
```diff
@@ -40,24 +41,39 @@ export async function installNodeModules(
...
-    const isYarn2 = await isUsingYarn2(stepCtx.workingDirectory);
-    if (isYarn2) {
-      args = ['install', '--no-immutable', '--inline-builds'];
-    }
+  let packageJson = {};
+  try {
+    packageJson = readPackageJson(stepCtx.workingDirectory);
+  } catch {
+    logger.info(
+      `Failed to read package.json. We won't know if we can use frozen lockfile. You can use EAS_NO_FROZEN_LOCKFILE=1 to disable it.`
```
brentvatne, line 49:
```suggestion
      `Failed to read package.json, defaulting to installing dependencies with a frozen lockfile. You can use EAS_NO_FROZEN_LOCKFILE=1 to disable it.`
```

packages/build-tools/src/utils/packageManager.ts
```diff
@@ -34,3 +35,59 @@ export async function isAtLeastNpm7Async(): Promise<boolean> {
...
+
+  if (sdkVersion && semver.lt(sdkVersion, '53.0.0')) {
+    // Before SDK 53 we could not have used frozen lockfile.
+    return false;
+  }
+
+  if (reactNativeVersion && semver.lt(reactNativeVersion, '0.79.0')) {
+    // Before react-native 0.79 we could not have used frozen lockfile.
+    return false;
+  }
```
brentvatne, line 60: not a blocker here but the dependence on reading a version from package.json feels like a bit of a ticking timebomb for us - at some point someone will need to use a fork and anything that depends on this type of logic will have undefined behavior. we may want to proactively figure out an alternative way to support this -- so if we detect you are using a fork of either package in your package.json then we let people know in eas cli or elsewhere that they need to specify the version explicitly for either or both of the packages somewhere in their package.json

brentvatne (approved):

packages/build-tools/src/common/prebuild.ts
```diff
@@ -32,9 +32,13 @@ export async function prebuildAsync<TJob extends BuildJob>(
...
   });
   const installDependenciesSpawnPromise = (
-    await installDependenciesAsync(ctx, {
+    await installDependenciesAsync({
+      packageManager: ctx.packageManager,
+      env: ctx.env,
       logger,
       cwd: resolvePackagerDir(ctx),
+      // prebuild sometimes modifies package.json, so we don't want to use frozen lockfile
+      useFrozenLockfile: false,
```
brentvatne, line 41: i believe it only modifies package.json if using a custom template, we tried to move away from modifying dependencies in prebuild in part to fix this problem of not being able to use frozen lockfiles. perhaps @byCedric has the most current context on this

brentvatne:
should we list this in docs on https://docs.expo.dev/eas/environment-variables/ or somewhere else that provides a reference for built-in EAS environment variables?

packages/build-tools/src/utils/packageManager.ts
sjchmiela: Totally agree. Added https://linear.app/expo/issue/ENG-15625/review-package-version-dependant-code-in-worker-and-think-of-better

sjchmiela:
New SDK53 project got `--frozen-lockfile` after these changes ([build](https://staging.expo.dev/accounts/sjchmiela/projects/staging-sdk-53/builds/a94dab85-9c1c-4b7e-80b8-0e3849d74682)). Old SDK52 project got no `--frozen-lockfile` ([build](https://staging.expo.dev/accounts/sjchmiela/projects/staging-app/builds/0ff8ad03-71a1-4dc0-bd20-4be5be7a847b)). So I guess it's as good to go as it can be? Will run the full system tests suite and deploy to production tomorrow CEST morning.

PR #543, opened 2025-06-12
Add an opt in env for deployment pod install

konrad-armatys:
# Why

This PR adds an option to use the `--deployment` flag on the `pod install` command.

This feature would:
- make the build fail on the pod install step rather than fingerprint check one
- give a more descriptive error in case user forgot to run `pod install`
- allow developers to use this stricter installation mode without requiring custom build configurations

[CocoaPods docs](https://guides.cocoapods.org/terminal/commands.html#pod_install)

# How

We read the feature flag `POD_INSTALL_DEPLOYMENT` and if it is set we add `--deployment` flag to `pod install` command

# Test Plan

Not needed

### DIFF
--- a/packages/build-tools/src/ios/pod.ts
+++ b/packages/build-tools/src/ios/pod.ts
@@ -12,7 +12,7 @@ export async function installPods<TJob extends Ios.Job>(
   const iosDir = path.join(ctx.getReactNativeProjectDirectory(), 'ios');
 
   const verboseFlag = ctx.env['EAS_VERBOSE'] === '1' ? ['--verbose'] : [];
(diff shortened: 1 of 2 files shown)


karmatys8:
> Should this be the default behavior in EAS?

Yes. I do not see any obvious cons of adding `--deployment`.
If you are okay with it, I can make the necessary changes. Should I also include an env allowing users to opt out of using this flag?

sjchmiela:
Let's _not_ make it the default. Then we're good to merge!

karmatys8:
Done.
Could you share the reason why you do not want to make it the default?

karmatys8:
@sjchmiela can you please merge this PR?

PR #555, opened 2025-06-24
[build-tools] Delete `.expo` directory when preparing project

sjchmiela:
# Why

https://exponent-internal.slack.com/archives/C02123T524U/p1750714495497629

# How

The prepare project phase is run for Android and iOS builds.

I'm not 100% convinced this is a good change. I worry down the build process some things may, for instance, expect the repository to be clean and, if `.expo` is being tracked, we'll make the working directory dirty. Isn't it better to fail here than there? What do you think?

# Test Plan

Added tests.


brentvatne:
maybe another alternative to this is to warn in doctor if `.expo` is included in your git repo. cc @keith-kurak

PR #583, opened 2025-08-26
[build-tools] Add `urql` and `gql.tada` to `build-tools`

sjchmiela:
# Why

So far we've been using HTTP endpoints for communication with `www`. Questions about whether we should move to GraphQL have been raised a couple times and the answer was always "it's a small API, we can make breaking changes on it easily, it's simpler".

Today I've come to upload project metadata file from here. Creating upload sessions exists in GraphQL, adding it to `www` would be re-adding `UploadSessionsRouter` removed not so long ago. Similarly, updating build metadata exists in GraphQL.

Maybe adding a GraphQL client to `build-tools` won't be so bad after all?

# How

Used the best GraphQL parsing engine and client there is, obviously.

I wanted to be able to switch between requesting GraphQL schema from localhost vs production easily, so I added `gql` and `gql:local` scripts. `yarn gql` runs now before `start`, `watch` and `build`.

Also integrated into CI.

# Test Plan

CI passes.

### DIFF
--- a/packages/build-tools/package.json
+++ b/packages/build-tools/package.json
@@ -21,7 +21,7 @@
     "jest-unit-watch": "jest --config jest/unit-config.ts --watch",
     "jest-integration-watch": "jest --config jest/integration-config.ts --watch",
     "clean": "rm -rf node_modules dist coverage",
-    "gql": "gql.tada generate-schema ${API_SERVER_URL:-https://api.expo.dev}/graphql --output ./schema.graphql",
+    "gql": "gql.tada generate-schema ${API_SERVER_URL:-https://api.expo.dev}/graphql --output ./schema.graphql && gql.tada generate-output",
     "gql:local": "API_SERVER_URL=http://api.expo.test yarn gql"
   },
   "author": "Expo <support@expo.io>",


sjchmiela:
wdyt?

packages/build-tools/package.json
```diff
@@ -11,14 +11,16 @@
   ],
   "scripts": {
     "start": "yarn watch",
-    "watch": "tsc --watch --preserveWatchOutput",
-    "build": "tsc",
+    "watch": "yarn gql && tsc --watch --preserveWatchOutput",
+    "build": "yarn gql && tsc",
```
kitten, line 15: nit: does it make sense to use `prewatch` & `prebuild`?

If we commit the `schema.graphql`, we could also choose not to commit `graphql-env.d.ts` and instead choose to use `gql.tada generate-output` in `build` instead

kitten (approved):
Seems fine 😅 There's a chance we'd have to add the plugin to the root `tsconfig.json` as well, since this repo doesn't use composites (which are also not supported), but havent't tested that yet, since the basic setup to write queries isn't added yet

PR #597, opened 2025-09-15
fix(build-tools): Add `--include=dev` to `npm ci`

kitten:
# Why

`npm ci` omits `devDependencies` by default when `NODE_ENV=production` is set. This is a discrepancy compared to all of our other installations.

https://docs.npmjs.com/cli/v10/commands/npm-ci

# How

- Switch to add `--include=dev` to `npm ci`

# Test Plan

- Created a new test project and ran `NODE_ENV=production npm ci --include=dev` (and without) to validate whether npm's docs are accurate

### DIFF
--- a/packages/build-tools/src/common/installDependencies.ts
+++ b/packages/build-tools/src/common/installDependencies.ts
@@ -25,7 +25,7 @@ export async function installDependenciesAsync({
   let args: string[];
   switch (packageManager) {
     case PackageManager.NPM: {
-      args = [useFrozenLockfile ? 'ci' : 'install'];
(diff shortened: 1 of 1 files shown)


sjchmiela (approved):
Oh right now I see [v9 documentation](https://docs.npmjs.com/cli/v9/commands/npm-ci) mentions `--include` even though it isn't in the table of contents (there's only `--omit` so I thought we may need to use one or the other depending on version).

NPM v8 mentions `--include` too. NPM v7 does not, but it's from 2020, Node 15. I think this is good. Sorry for false request-changes!

PR #598, opened 2025-09-15
[build-tools] Support yarn workspaces focus to install scoped dependencies

sebryu:
# Why

I have monorepo with multiple expo apps (+ web apps). I want to use `yarn workspaces focus my-app`, so node_modules is slimmer and yarn install faster ([see workspaces focus](https://yarnpkg.com/cli/workspaces/focus)). 

By using workspaces focus I would only install dependencies used by a single app, and without this change on EAS I would get inconsistent fingerprint between local and server envs.

# How

Added custom flag which would override install dependencies command, so by running:
```
EAS_YARN_FOCUS_WORKSPACE='my-app` eas build
```
The dependencies would be installed with `yarn workspaces focus my-app` command.

# Test Plan

I have run eas-build locally with those changes.

### DIFF
--- a/packages/build-tools/src/common/installDependencies.ts
+++ b/packages/build-tools/src/common/installDependencies.ts
@@ -36,7 +36,7 @@ export async function installDependenciesAsync({
       const isModernYarnVersion = await isUsingModernYarnVersion(cwd);
       if (isModernYarnVersion) {
         if (env['EAS_YARN_FOCUS_WORKSPACE']) {
-          args = ['workspaces', 'focus', env['EAS_YARN_FOCUS_WORKSPACE'], '--production'];
+          args = ['workspaces', 'focus', env['EAS_YARN_FOCUS_WORKSPACE']];
         } else {
           args = ['install', '--inline-builds', useFrozenLockfile ? '--immutable' : '--no-immutable'];
         }


packages/build-tools/src/common/installDependencies.ts
```diff
@@ -47,6 +47,10 @@ export async function installDependenciesAsync({
     default:
       throw new Error(`Unsupported package manager: ${packageManager}`);
   }
+  if (env['EAS_YARN_FOCUS_WORKSPACE']) {
+    // override package manager to yarn if this flag is set
+    [packageManager, ...args] = env['EAS_YARN_FOCUS_WORKSPACE'].split(' ') as [PackageManager, ...string[]];
+  }
```
sjchmiela, line 53: how about sth like


```suggestion
  if (env['EAS_YARN_FOCUS_WORKSPACE']) {
    if (packageManager != yarn) {
      logger.warn("EAS_YARN_FOCUS_WORKSPACE is set, but the workspace is not using yarn.")
    } else {
      logger.info(`EAS_YARN_FOCUS_WORKSPACE is set to '${...}', changing install command to ...`);
      args = ["workspaces", "focus", env['EAS_YARN_FOCUS_WORKSPACE']];
    }
  }
```
?

PR #599, opened 2025-09-15
[ios-builder] support save and restore ccache for builds and workflows

AbbanMustafa:
# Why

Provide support for utilizing ccache with the recently added `save` and `restore` cache functions for ios Builds and Workflows. This is intended to reduce job times by relying on our provided cache for redundant work, subsequently leading to efficient jobs and queues.

# How
A cache router has been defined for both builds and workflows. The caching steps are only enabled for builds that run on our workers, denoted with EAS_BUILD_RUNNER. The worker sets relevant environment variables that enables Fastlane to apply ccache https://github.com/expo/turtle-v2/pull/2398

# Test Plan
Validated with local build and workflow runs. Observed Fastlane runtime improvement from 3min to ~50 seconds, ensured ccache was successfully restored, used, and saved
Screenshot 2025-09-22 at 10 50 41 PM
Restore step matches key based off package manager diff and checks against cache route in www 
Screenshot 2025-09-23 at 12 52 24 AM
Save performs the same diff 
Screenshot 2025-09-23 at 12 53 10 AM

```
➜  ccache -sv
Cache directory:    /Users/mustafa/Library/Caches/ccache
Config file:        /Users/mustafa/Library/Preferences/ccache/ccache.conf
System config file: /opt/homebrew/etc/ccache.conf
Stats updated:      never
Local storage:
  Cache size (GiB): 0.0 / 5.0 ( 0.00%)
  Files:              0
  Hits:               0
  Misses:             0
  Reads:              0
  Writes:             0
➜  ccache -sv
Cache directory:      /Users/mustafa/Library/Caches/ccache
Config file:          /Users/mustafa/Library/Preferences/ccache/ccache.conf
System config file:   /opt/homebrew/etc/ccache.conf
Stats updated:        Mon Sep 22 21:12:43 2025
Cacheable calls:      10128 / 10136 (99.92%)
  Hits:                7596 / 10128 (75.00%)
    Direct:            7596 /  7596 (100.0%)
    Preprocessed:         0 /  7596 ( 0.00%)
  Misses:              2532 / 10128 (25.00%)
Uncacheable calls:        8 / 10136 ( 0.08%)
  Called for linking:     8 /     8 (100.0%)
Successful lookups:
  Direct:              7596 / 10128 (75.00%)
  Preprocessed:           0
Local storage:
  Cache size (GiB):     0.9 /   5.0 (18.44%)
  Files:               7596
  Hits:                7596 / 10128 (75.00%)
  Misses:              2532 / 10128 (25.00%)
  Reads:              17724
  Writes:              5064
```

### DIFF
--- a/packages/build-tools/src/builders/ios.ts
+++ b/packages/build-tools/src/builders/ios.ts
@@ -78,7 +78,7 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
       });
     });
 
-    if (ctx.env.EAS_BUILD_RUNNER == 'eas-build') {
+    if (ctx.env.EAS_BUILD_RUNNER === 'eas-build') {
       await ctx.runBuildPhase(BuildPhase.RESTORE_CACHE, async () => {
         const workingDirectory = ctx.getReactNativeProjectDirectory();
         const paths = [path.join(ctx.env.HOME, 'Library/Caches/ccache')];
@@ -95,26 +95,29 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
         } catch (err) {
           ctx.logger.warn({ err }, 'Failed to read package files for cache key generation');
         }
-        ctx.logger.info("RESTORE - ", keyData)
+        ctx.logger.info('RESTORE - ', keyData);
         const cacheKey = `ios-cache-${createHash('sha256').update(keyData).digest('hex').substring(0, 16)}`;
 
         let jobRunId;
-        if(ctx.env.EAS_BUILD_ID){
-          jobRunId = ctx.env.EAS_BUILD_ID
+        if (ctx.env.EAS_BUILD_ID) {
+          jobRunId = ctx.env.EAS_BUILD_ID;
         } else {
-          jobRunId = nullthrows(ctx.env.__WORKFLOW_JOB_ID, 'EAS_BUILD_ID or __WORKFLOW_JOB_ID is not set');
+          jobRunId = nullthrows(
+            ctx.env.__WORKFLOW_JOB_ID,
+            'EAS_BUILD_ID or __WORKFLOW_JOB_ID is not set'
+          );
         }
         const robotAccessToken = nullthrows(
           ctx.job.secrets?.robotAccessToken,
           'Robot access token is required for cache operations'
         );
 
-        ctx.logger.info("key - ", cacheKey)
-        ctx.logger.info("paths - ", paths)
+        ctx.logger.info('key - ', cacheKey);
+        ctx.logger.info('paths - ', paths);
         try {
           const { archivePath } = await downloadCacheAsync({
             logger: ctx.logger,
-            buildId:jobRunId,
+            buildId: jobRunId,
             expoApiServerURL: ctx.env.__API_SERVER_URL ?? 'https://exp.host',
             robotAccessToken,
             paths,
@@ -135,7 +138,7 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
           ctx.logger.warn({ err }, 'Failed to restore cache');
         }
       });
-  }
+    }
 
     await ctx.runBuildPhase(BuildPhase.INSTALL_PODS, async () => {
       await runInstallPodsAsync(ctx);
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/steps/functions/saveCache.ts
```diff
@@ -76,29 +73,39 @@ export function createSaveCacheFunction(): BuildFunction {
 
 export async function uploadCacheAsync({
   logger,
-  jobRunId,
+  buildId,
```
sjchmiela, line 76: Replacing `jobRunId` with `buildId` makes it weird for job runs? For single arguments like these I've been using `taskId` (from lack of better options). Maybe `jobId` would be good too.

We can also consider passing in a `ctx: OneOfBuildContexts` and let `uploadCacheAsync` figure out:
- whether we're in a build or a job run
- robot access token
- the path to use
etc.

sjchmiela (changes requested):

packages/build-tools/src/builders/ios.ts
```diff
@@ -170,9 +215,49 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
     });
   });
 
-  await ctx.runBuildPhase(BuildPhase.SAVE_CACHE, async () => {
-    await ctx.cacheManager?.saveCache(ctx);
-  });
+  if (ctx.env.EAS_BUILD_RUNNER === 'eas-build') {
```
sjchmiela, line 218: Let's have the build phase run and skip the implementation if EAS_BUILD_RUNNER !== eas-build. This will also deindent the code because you can early return instead of wrapping big chunk of code with `if`

packages/build-tools/src/builders/ios.ts
```diff
@@ -170,9 +215,49 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
...
+        );
+        const expoApiServerURL = nullthrows(
+          ctx.env.__API_SERVER_URL,
+          '__API_SERVER_URL is not set'
+        );
+
+        const { archivePath } = await compressCacheAsync({
+          paths,
+          workingDirectory,
+          verbose: true,
```
sjchmiela, line 239: Only if EXPO_DEBUG is truthy?

packages/build-tools/src/builders/ios.ts
```diff
@@ -170,9 +215,49 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
...
+          expoApiServerURL,
+          robotAccessToken,
+          archivePath,
+          key: cacheKey,
+          paths,
+          size,
+          platform: ctx.job.platform,
+        });
+      } catch (err) {
+        ctx.logger.warn({ err }, 'Failed to save cache');
```
sjchmiela, line 257: I think we can `.error` here, why not.

AbbanMustafa:
With eviction
Before
Screenshot 2025-09-25 at 5 32 14 PM
After
Screenshot 2025-09-25 at 5 31 35 PM

packages/build-tools/src/builders/ios.ts
```diff
@@ -73,7 +80,47 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
     });
 
     await ctx.runBuildPhase(BuildPhase.RESTORE_CACHE, async () => {
-      await ctx.cacheManager?.restoreCache(ctx);
+      if (ctx.env.EAS_USE_CACHE !== '1') {
+        return;
+      }
+      const workingDirectory = ctx.getReactNativeProjectDirectory();
+      const paths = [path.join(ctx.env.HOME, 'Library/Caches/ccache')];
```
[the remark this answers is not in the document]
AbbanMustafa: good point, moved them up

packages/build-tools/src/builders/ios.ts
```diff
@@ -171,7 +218,47 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
...
+
+      const robotAccessToken = nullthrows(
+        ctx.job.secrets?.robotAccessToken,
+        'Robot access token is required for cache operations'
+      );
+      const expoApiServerURL = nullthrows(ctx.env.__API_SERVER_URL, '__API_SERVER_URL is not set');
+
+      // cache size can blow up over time over many builds, so evict stale files and only upload what was used within this builds time window
+      const evictWindow = Math.floor((Date.now() - buildStart) / 1000);
+      await spawnAsync('ccache', ['--evict-older-than', evictWindow + 's']);
```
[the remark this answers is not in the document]
AbbanMustafa: thank you yes logging provides great detail
Screenshot 2025-09-26 at 4 15 30 PM

AbbanMustafa:
on no cache found (turtlefetch download returns 404)
Screenshot 2025-09-26 at 5 02 03 PM

PR #600, opened 2025-09-17
[build-tools] hashFiles interpolation function

AbbanMustafa:
# Why
Provide an interpolation function `hashFiles` that returns a single hash for the set of files that matches the path pattern. Users can provide a single path pattern or multiple path patterns separated by commas. 
```
key: key-${{ hashFiles('package-lock.json') }}
key: key-${{ hashFiles('*.json', '!package-lock.json') }}
```

# How
  - Implements hashFiles() function for dynamic cache key generation (following GitHub Actions)
  - Adds support for multiple glob patterns in hashFiles()
# Test Plan
Added relevant tests to BuildStepContext, and validated cache key with local workflow. Ran scenarios to ensure file changes propagate with new cache keys, invalid files result in empty keys, multiple and exclusion filters correctly
```
- uses: eas/restore_cache
  with:
    path: "/Users/mustafa/Desktop/foo"
    key: key-${{ hashFiles('package-lock.json') }}
- run: sleep 10
- uses: eas/save_cache
  with:
    path: /Users/mustafa/Desktop/foo
    key: key-${{ hashFiles('package-lock.json') }}
```
Screenshot 2025-10-07 at 6 30 56 PM

```
  - uses: eas/save_cache
    with:
      path: /Users/mustafa/Desktop/foo
      key: key-${{ hashFiles('package-lock-foo.json') }}
```
Screenshot 2025-10-07 at 6 36 16 PM

exclusion
```
- uses: eas/save_cache
  with:
    path: /Users/mustafa/Desktop/foo
    key: key-${{ hashFiles('!package-lock.json') }}
```
Screenshot 2025-10-07 at 7 44 14 PM

### DIFF
--- a/packages/steps/src/BuildStepContext.ts
+++ b/packages/steps/src/BuildStepContext.ts
@@ -199,7 +199,7 @@ export class BuildStepGlobalContext {
     }
 
     const result = crypto.createHash('sha256');
-    let count = 0;
+    let fileFound = false;
 
     for (const file of filePaths) {
       if (!file.startsWith(`${workspacePath}${path.sep}`)) {
@@ -214,12 +214,12 @@ export class BuildStepGlobalContext {
       const fileContent = fs.readFileSync(file);
       fileHash.update(fileContent);
       result.write(fileHash.digest());
-      count++;
+      fileFound = true;
     }
 
     result.end();
 
-    if (count > 0) {
+    if (fileFound) {
       return result.digest('hex');
     } else {
       return '';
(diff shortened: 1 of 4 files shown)


packages/steps/src/BuildStepContext.ts
```diff
@@ -180,6 +184,67 @@ export class BuildStepGlobalContext {
...
+        }
+        hash.update('\0'); // separator between files
+      }
+
+      return hash.digest('hex');
+    } catch (error: any) {
+      // Return a consistent hash for errors
+      const errorHash = crypto.createHash('sha256');
+      errorHash.update(`ERROR: ${error?.message ?? 'Unknown error'}`);
+      return errorHash.digest('hex');
```
sjchmiela, line 244: Why return a hash in case of an error? I think we should just error?

sjchmiela (changes requested):

PR #614, opened 2025-10-06
save/restore flag, prefix, cache stats

AbbanMustafa:
# Why

Please describe the motivation for this PR, and link to relevant GitHub issues, forums posts, or feature requests.

# How

How did you build this feature or fix this bug and why?

# Test Plan

Screenshot 2025-10-06 at 2 19 13 PM

Screenshot 2025-10-06 at 2 53 05 PM

### DIFF
--- a/packages/build-tools/src/builders/android.ts
+++ b/packages/build-tools/src/builders/android.ts
@@ -33,6 +33,8 @@ import { findPackagerRootDir } from '../utils/packageManager';
 import { runBuilderWithHooksAsync } from './common';
 import { runCustomBuildAsync } from './custom';
 
+const CACHE_KEY_PREFIX = 'android-ccache'
+
 export default async function androidBuilder(ctx: BuildContext<Android.Job>): Promise<Artifacts> {
   if (ctx.job.mode === BuildMode.BUILD) {
     await prepareExecutableAsync(ctx);
@@ -74,7 +76,7 @@ async function buildAsync(ctx: BuildContext<Android.Job>): Promise<void> {
   });
 
   await ctx.runBuildPhase(BuildPhase.RESTORE_CACHE, async () => {
-    if (ctx.env.EAS_USE_CACHE !== '1') {
+    if (ctx.env.EAS_USE_CACHE !== '1' && ctx.env.EAS_RESTORE_CACHE !=='1') {
       // EAS_USE_CACHE is for the new cache. If it is not set, use the old cache.
       await ctx.cacheManager?.restoreCache(ctx);
       return;
@@ -97,7 +99,7 @@ async function buildAsync(ctx: BuildContext<Android.Job>): Promise<void> {
         robotAccessToken,
         paths: cachePaths,
         key: cacheKey,
-        keyPrefixes: [],
+        keyPrefixes: [CACHE_KEY_PREFIX],
         platform: ctx.job.platform,
       });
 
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/builders/android.ts
```diff
@@ -74,7 +76,7 @@ async function buildAsync(ctx: BuildContext<Android.Job>): Promise<void> {
   });
 
   await ctx.runBuildPhase(BuildPhase.RESTORE_CACHE, async () => {
-    if (ctx.env.EAS_USE_CACHE !== '1') {
+    if (ctx.env.EAS_USE_CACHE !== '1' && ctx.env.EAS_RESTORE_CACHE !== '1') {
```
sjchmiela, line 79: Actually there's one more change we should do: run `await ctx.cacheManager?.restoreCache(ctx);` always (we never said we're turning off the old cache).

Also, can we please rewrite the logic to something more verbose, but more easily understandable? What I mean is that
```
  (USE_CACHE == 1 && RESTORE_CACHE != 0) // we want to use cache, but are forbidden from restoring
  ||
  (RESTORE_CACHE == 1) // we want to restore no matter what
```
is more readable to me as a human than two AND-ed negations. Are these two equivalent?

packages/build-tools/src/builders/android.ts
```diff
@@ -33,6 +33,8 @@ import { findPackagerRootDir } from '../utils/packageManager';
 import { runBuilderWithHooksAsync } from './common';
 import { runCustomBuildAsync } from './custom';
 
+const CACHE_KEY_PREFIX = 'android-ccache';
```
sjchmiela, line 36:
```suggestion
const CACHE_KEY_PREFIX = 'android-ccache-';
```

packages/build-tools/src/builders/android.ts
```diff
@@ -300,7 +313,7 @@ async function generateCacheKeyAsync(workingDirectory: string): Promise<string>
 
   try {
     const key = await hashFiles([lockPath]);
-    return `android-ccache-${key}`;
+    return `${CACHE_KEY_PREFIX}-${key}`;
```
sjchmiela, line 316:
```suggestion
    return `${CACHE_KEY_PREFIX}${key}`;
```

Same changes for ios

packages/build-tools/src/builders/android.ts
```diff
@@ -109,8 +111,15 @@ async function buildAsync(ctx: BuildContext<Android.Job>): Promise<void> {
...
 
       ctx.logger.info('Cache restored successfully');
-    } catch (err) {
-      ctx.logger.warn({ err }, 'Failed to restore cache');
+      await asyncResult(
+        spawnAsync('ccache', ['--zero-stats'], { env: ctx.env, logger: ctx.logger, stdio: 'pipe' })
+      );
+    } catch (err: any) {
+      if (err.response.status === 404) {
+        ctx.logger.warn('No cache found for this key, ensure it was created with eas/save_cache');
```
sjchmiela, line 119: I think we want to `info`.

Also, actually, maybe the action for users here is not to ensure, but rather to… just build haha (let's not ask them to ensure anything)

packages/build-tools/src/builders/android.ts
```diff
@@ -109,8 +111,15 @@ async function buildAsync(ctx: BuildContext<Android.Job>): Promise<void> {
...
+      } else {
+        ctx.logger.warn({ err }, 'Failed to restore cache');
```
sjchmiela, line 121: 👍

packages/build-tools/src/builders/android.ts
sjchmiela: We can also check what does github actions print in their cache action

sjchmiela: Though I think our message may be even less actionable than theirs given our cache integration is much more integrated and we can hopefully expect like 90% of users to just never disable cache and be happy with the results of what we're devising here, whereas in GHA it's all manual

